PR #2, opened 2018-10-26
[PCM-731] The streaming client will request a channel using the strea…

xdumaine:
…ming connection type and use it's id in the channel auth


test/unit/client.test.js
```diff
@@ -2,143 +2,107 @@
```
xdumaine: Big test refactor to make them actually have value.

PR #8, opened 2018-12-11
Add docs, make api methods return promises

xdumaine:
# 8.0.0

Added:
- documentation 🎉 
- bulkSubscribe for notifications to reduce requests for subscribing to lots of topics at once

Breaking changes:
- default export is a constructor instead of a named function exported
- the constructor has `extend` as a static method
- extension API changed to take the client instance as a first parameter instead of the underlying stanzaio instance. This is to provide easier access to config options and event emitting/listening to extensions. 
- notifications.subscribe, notifications.unsubscribe, client.connect, client.disconnect, client.reconnect all now return promises for easier async usage (previous required added event listeners separate from the calls)


doc/documentation.md
```diff
@@ -0,0 +1,87 @@
...
+#### Constructor
+
+`new PureCloudStreamingClient(options)`
+
+- parameters
+  - `Object options` with properties:
+    - `String authToken`: Required; access token for the user
+    - `String host`: Required; `wss://streaming.` + `mypurecloud.com || mypurecloud.ie ||
+        mypurecloud.jp || mypurecloud.de || mypurecloud.com.au`
+    - `String jid` : Required; JabberId for the user (get `from api/v2/users/me`)
```
jensengar: from should not be in the back ticks.

doc/documentation.md
```diff
@@ -0,0 +1,87 @@
...
+
+#### Methods
+
+`client.connect() : Promise<void>` - Initialize the WebSocket connection for streaming
+connectivity with PureCloud. Initialize must be called before any events will trigger.
```
jensengar: I don't see `Initialize` as a method or anything, did you mean `connect`?

ckbrewer33: I think the method _is_ `connect`, the description should probably say 'Initializes the WebSocket...'

doc/documentation.md
```diff
@@ -0,0 +1,87 @@
...
+
+`client.reconnect() : Promise<void>` - Disconnect (if connected) and reconnect to
+the streaming service
+
+`client.disconnect() : Promise<void>` - Disconnect from  the streaming
```
ckbrewer33: remove extra space between 'from' and 'the'

doc/documentation.md
```diff
@@ -0,0 +1,87 @@
...
+service
+
+`client.on(eventName, handler) : void` - register an event handler for the client
+
+- parameters
+  - `String eventName` - event name to watch
```
ckbrewer33: Is there a specific set of event names that we could list?

doc/notifications.md
```diff
@@ -0,0 +1,47 @@
...
+
+`client.notifications.subscribe(topic, handler) : Promise<void>` - For registering
+a handler for a topic
+
+`client.notifications.unsubscribe(topic, handler) : Promise<void>` For unregistering
+a handler for a topic
+
+When unsubscribing, either through bulk or individual subscriptions, the client will
+only unsubscribe the topic from the backend if there are no more handlers. Since bulk
+subscriptions don't have a specific handler for each topic, they could as a single "handler"
```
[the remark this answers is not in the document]
xdumaine: `could` -> `count`. Took me a minute to figure this out.

PR #15, opened 2019-02-22
[PCM-830] Prevent need for dummy function to subscribe

xdumaine:


test/unit/notifications.test.js
```diff
@@ -119,6 +119,10 @@ test('subscribe and unsubscribe do their jobs', async t => {
   // there are still more subscriptions
   sinon.assert.notCalled(notification.client._stanzaio.unsubscribeFromNode);
 
+  await notification.expose.unsubscribe('test', () => {});
+  // unsubscribing with an unused handler won't trigger any unsubscribe
+  sinon.assert.notCalled(notification.client._stanzaio.unsubscribeFromNode);
+
   await notification.expose.unsubscribe('test');
   // unsubscribing without a handler won't trigger any unsubscribe
   sinon.assert.notCalled(notification.client._stanzaio.unsubscribeFromNode);
```
xdumaine: This is no bueno. Need to fix ensure unsubscribe does in fact unsubscribe. This might just be test weirdness.

PR #26, opened 2019-07-12
Upgrade more deps

xdumaine:
Needs close review of everything outside package-lock.json


xdumaine:
Will be major version bump. Requires major version bump of webrtc-sessions before this cam be merged.

PR #28, opened 2019-07-22
[PCM-1018] Fix reconnect behavior to reconnect on unauthorized after …

xdumaine:
…an initial connection has been made

### DIFF
--- a/src/reconnector.js
+++ b/src/reconnector.js
@@ -39,19 +39,25 @@ class Reconnector {
     });
 
     this.backoff.on('fail', () => {
-      this.client.logger.error('Failed to reconnect to the streaming service. Attempting to connect with new channel.');
-      // attempt with a new channel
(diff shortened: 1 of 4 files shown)


src/reconnector.js
```diff
@@ -39,19 +39,24 @@ class Reconnector {
...
+    this.client.on('connected', () => {
+      this._hasConnected = true;
+      this._cleanupReconnect();
+    });
 
     // disable reconnecting when there's an auth failure
     this.client.on('sasl:failure', (err) => {
       if (!err || err.condition !== 'temporary-auth-failure') {
+        // an auth failure when it's already connected once means channel expiry
+        if (err && err.condition === 'not-authorized' && this._hasConnected) {
```
ckbrewer33: NIT:  Pull out this logic into a variable called `channelExpired` or something like that and you won't need the comment

PR #34, opened 2019-10-03
[PCM-1101] Add automatic combining of topics when subscribing

xdumaine:
~~This intentionally does not do combining of topics when registering individual topics.~~ When doing both subscribe or bulkSubscribe, it does it for you. This means that all consumers get this new behavior for free, and effectively get an increase in their max allowable topics, since many will be combined.


src/notifications.js
```diff
@@ -71,12 +71,49 @@ export default class Notification {
...
+    let combinedTopics = [];
+
+    // Max length of 200 in topic names
+    // so recursively break them up if the combined length exceeds 200
+    const combineTopics = function (prefix, postFixes) {
+      const delimiter = postFixes.length === 1 ? '.' : '?';
+      const id = `${prefix}${delimiter}${postFixes.join('&')}`;
+      if (id.length < 200) {
+        combinedTopics.push({ id });
+      } else {
```
jensengar: Would it ever be possible to have id.length >= 200 and only one postfix? if so, this would loop forever right?

xdumaine: Fixed

xdumaine: Fixed

PR #36, opened 2019-10-15
Updated webpack build to include babel's transform-runtime instead of the full babel-polyfill

djhouseknecht:
TLDR: use `'@babel/plugin-transform-runtime'` instead of `'babel-polyfill'`

Long story: `babel-polyfill` can only be loaded once per browser page. Ie, if two libs or apps try to load it in the same page, it will throw an error (I don't love that implementation). But, babel and their infinite config options has a `transform-runtime` plugin that will essentially bundle the needed parts of the polyfill into the lib/app. This way other applications can use streaming-client without the `babel-polyfill` collision.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -38,10 +38,12 @@ module.exports = (env) => {
     module: {
       rules: [
         {
+          exclude: /(node_modules)/,
           test: /\.js$/,
           loader: 'babel-loader',
           query: {
-            presets: ['@babel/preset-env']
+            presets: ['@babel/preset-env'],
+            plugins: ['@babel/plugin-transform-runtime']
           }
         }
       ]
--- a/package-lock.json
+++ b/package-lock.json
@@ -1268,6 +1268,18 @@
         "@babel/helper-plugin-utils": "^7.0.0"
       }
     },
+    "@babel/plugin-transform-runtime": {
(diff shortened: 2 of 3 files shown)


webpack.config.js
```diff
@@ -38,10 +38,12 @@ module.exports = (env) => {
     module: {
       rules: [
         {
+          exclude: /(node_modules)/,
           test: /\.js$/,
           loader: 'babel-loader',
           query: {
-            presets: ['@babel/preset-env']
+            presets: ['@babel/preset-env'],
+            plugins: ['@babel/plugin-transform-runtime']
```
ckbrewer33: What does this do that the poly-fill wansn't?

djhouseknecht: https://babeljs.io/docs/en/babel-plugin-transform-runtime#why

Main reason is this will not pollute the global namespace. 
> Another purpose of this transformer is to create a sandboxed environment for your code. If  you directly import `core-js` or `@babel/polyfill` and the built-ins it provides such as `Promise`, `Set` and `Map`, those will pollute the global scope. While this might be ok for an app or a command line tool, it becomes a problem if your code is a library which you intend to publish for others to use or if you can't exactly control the environment in which your code will run.

PR #40, opened 2019-11-04
[PCM-1144] no_longer_subscribed should trigger a new channel

jensengar:


src/client.js
```diff
@@ -140,7 +166,7 @@ export default class Client {
         // default rate limit
         // 20 stanzas per 1000 ms,
         // adding up to 25 stanzas over the course of the 1000ms
-        // starting with 20 stanzas
+        // starting th 20 stanzas
```
ckbrewer33: intentional?

src/client.js
```diff
@@ -123,7 +130,26 @@ export default class Client {
...
+        clearInterval(this.leakyReconnectTimer);
+        return;
+      }
+
+      this.hardReconnectCount++;
+
+      this.logger.info('no_longer_subscribed received, attempting hard reconnect');
+
+      if (!this.leakyReconnectTimer) {
+        this.leakyReconnectTimer = setInterval(() => {
```
ckbrewer33: Seems to me we should probably call `stopInterval()` at some point so it's not just sitting there counting in the background when the `hardReconnectCount` is 0

src/client.js
jensengar: no

src/client.js
jensengar: done

src/client.js
```diff
@@ -121,9 +137,38 @@ export default class Client {
       }
     });
 
-    this.on('no_longer_subscribed', (event) => {
+    this.on('notify:v2.system.no_longer_subscribed*', (event, data) => {
```
ckbrewer33: Where did you find the new event name?  Can we get it documented somewhere?

djhouseknecht: Could you explain the difference between a `no_longer_subscribed` event and a `_disconnected`? My understand is a `_disconnected` is when the WS closes/disconnects (could be network, etc). The `no_longer_subscribed` is when the API stops sending messages to the WS because the client reached their 10 limit. Does that sound right? Also, does the backend not close the WS when it sends a `no_longer_subscribed`? If it does, would it call `_disconnected` and `no_longer_subscribed` (which would explain why you have the `this.deadChannels` check in `on('_disconnected')`).

jensengar: so `no_longer_subscribed` is a message sent by the notifications service saying we've hit our channel limit and this channel is not longer valid. It then closes the websocket remotely (hence the deadChannels like you said). `_disconnected` will usually try and reconnect, but it will use the same channelId so we need to create a new channel and connect using that. The difference between `_disconnected` and `no_longer_subscribed` is `_disconnected` is the result of the websocket *actually* being disconnected, where `no_longer_subscribed` is just a message that comes across the channel. The `no_longer_subscribed` message *only* happens if you hit the limit where `_disconnected` happens every time the websocket closes. I hope that makes sense.

PR #42, opened 2019-11-20
Added changelog file

ckbrewer33:
With entries for v11.x.x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,11 @@
+#Changelog
+
+## v11.0.1
+* Added changelog file
+
+## v11.0.0
+* Added a new function to send an 'accept' event
+* No longer sending 'proceed' and 'accept' in the same call to `acceptRtcSession`
+
+#### Breaking Changes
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
...
+
+## v11.0.1
+* Added changelog file
+
+## v11.0.0
+* Added a new function to send an 'accept' event
+* No longer sending 'proceed' and 'accept' in the same call to `acceptRtcSession`
+
+#### Breaking Changes
+* In previous versions, calling `acceptRtcSession` on the `webrtcsessions` extension would send both a 'proceed' and 'accept' event in one call.  Beginning in version 11, the 'accept' has been moved to a separate function.  To do both a 'proceed' and an 'accept' you will need to call `webrtcsessions.acceptRtcSession` to send the proceed, and `webrtcsessions.rtcSessionAccepted` to send the accept.
```
jensengar: I'd provide a little more information here. Something like "This caused race conditions with multiple clients using autoAnswer where they would both call `acceptRtcSession` at the same time and send each other an `accept` stanza, resulting in neither client actually handling the incoming session. As a result, we recommend *not* calling `rtcSessionAccepted` until the session has actually been received.

PR #43, opened 2019-12-05
Don't automatically combine already combined topics

caitlanspronk:


src/notifications.js
```diff
@@ -77,7 +77,14 @@ export default class Notification {
 
   mapCombineTopics (topics) {
     const prefixes = {};
-    topics.map(t => {
+    const precombinedTopics = [];
+    topics.filter(t => {
+      if (t.indexOf('?') === -1) {
+        return true;
+      } else {
+        precombinedTopics.push(t);
```
jensengar: Also, this is a side-effect and should generally be avoided. By that, I mean you shouldn't do 
```
something.filter((item) => {
  if (!item.condition) {
    // do something unrelated
  }

  return item.condition
}
```
I understand *why* you are doing this and that it saves a line or two of code, but I'd prefer either using a forEach to create 2 different arrays, or running a separate dedicated filter to find the precombinedTopics. Does that make sense what I'm saying?

test/unit/notifications.test.js
```diff
@@ -370,3 +370,26 @@ test('notifications | mapCompineTopics should correctly reduce topics', t => {
   t.is(reducedTopics[1].id, 'v2.users.testuser.thisIsAReallyLongTopicForThePurposeOfExceeding200CharsinCombinedTopicNames');
   t.is(reducedTopics[2].id, 'v2.users.testuser?InRealityTheseWouldBeALotOfDisparateTopicsThatWhenJoinedExceed200Chars&athirdreallylongtopicathirdreallylongtopicathird');
 });
+
+test('notifications | mapCompineTopics should not combine already combined topics', t => {
```
jensengar: great work

src/notifications.js
caitlanspronk: Yep, it does, and will do!

src/notifications.js
```diff
@@ -77,7 +77,14 @@ export default class Notification {
 
   mapCombineTopics (topics) {
     const prefixes = {};
-    topics.map(t => {
+    const precombinedTopics = [];
+    topics.filter(t => {
+      if (t.indexOf('?') === -1) {
```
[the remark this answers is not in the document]
caitlanspronk: Okay--I since this is a public project I thought maybe we still wanted to try to be compatible, but I'm certainly happy to use includes instead :)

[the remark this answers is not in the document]
JackieChiles: True, though I'm willing to say dropping IE support extends to this library as well, I'm cool either way :+1:

PR #44, opened 2019-12-10
PCM-849 - Log Channel id with streaming client errors

ckbrewer33:
Added channelId to error logs
updated changelog file to include topic combining fix that went out in v11.0.1 release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 #Changelog
+## v11.0.2
+* Included channelId in error logs
 
 ## v11.0.1
 * Added changelog file
+* Fixed an issue with topics not being combined correctly - [PR #43](https://github.com/purecloudlabs/purecloud-streaming-client/pull/43)
 
 ## v11.0.0
 * Added a new function `rtcSessionAccepted` to send an 'accept' event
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -130,7 +130,8 @@ export default class Client {
     });
 
     this.on('sasl:failure', (err) => {
-      this.logger.error('Authentication failed connecting to streaming service', err);
+      const errMessage = `Authentication failed connecting to streaming service\nchannelId: ${this.config.channelId}`;
```
djhouseknecht: It looks like a minor typo in `service\nchannelId:` to `service\channelId:`.

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 #Changelog
+## v11.0.2
+* Included channelId in error logs
 
 ## v11.0.1
 * Added changelog file
+* Fixed an issue with topics not being combined correctly - [PR #43](https://github.com/purecloudlabs/purecloud-streaming-client/pull/43)
```
jensengar: thanks for putting this in.

src/client.js
ckbrewer33: That's a newline character :)

djhouseknecht: Oof, you are right. For some reason I was reading it like you were trying to do a normal `\`

PR #47, opened 2020-03-02
[PCM-1180] option to disable reconnect on no_longer_subscribed

jensengar:


src/client.js
```diff
@@ -102,6 +105,14 @@ export default class Client {
       this.connected = false;
       this._ping.stop();
 
+      if (!event) {
```
djhouseknecht: When does the streaming-client disconnect without emitting an event? Just for my knowledge.

jensengar: I honestly have no idea. I've seen it happen in the spigot tests. It originates in the stanza code: https://github.com/legastero/stanza/blob/f8638c4e670c8111599481d6eb36646bd76d3078/src/Client.ts#L253

jensengar: I assume this happens if you gracefully disconnect, where as the event only happens if the socket closes unexpectedly.

PR #52, opened 2020-04-07
Engageui 3795 topic limit truncation

caitlanspronk:


src/notifications.js
```diff
@@ -122,7 +123,21 @@ export default class Notification {
...
+  }
+
+  truncateTopicList (topics) {
+    const keptTopics = topics.slice(0, MAX_SUBSCRIBABLE_TOPICS);
+    if (topics.length > MAX_SUBSCRIBABLE_TOPICS) {
+      let droppedTopics = topics.slice(MAX_SUBSCRIBABLE_TOPICS);
+      if (droppedTopics.length > 20) {
+        const length = droppedTopics.length - 20;
+        droppedTopics = droppedTopics.slice(20);
+        droppedTopics.push(`...and ${length} more`);
```
jensengar, line 136: I don't love this, but I assume this is to make it so the log message isn't to big to be sent?

src/notifications.js
```diff
@@ -122,7 +123,21 @@ export default class Notification {
...
 
-    return combinedTopics.concat(precombinedTopics);
+    return this.truncateTopicList(combinedTopics.concat(precombinedTopics));
+  }
+
+  truncateTopicList (topics) {
+    const keptTopics = topics.slice(0, MAX_SUBSCRIBABLE_TOPICS);
+    if (topics.length > MAX_SUBSCRIBABLE_TOPICS) {
+      let droppedTopics = topics.slice(MAX_SUBSCRIBABLE_TOPICS);
+      if (droppedTopics.length > 20) {
```
jensengar, line 133: Can we get a var for this please? Something like `DROPPED_TOPICS_DISPLAY_COUNT` or something along those lines.

src/notifications.js
caitlanspronk: I assume so. I basically copied the existing behavior in EPC, but I don't have strong feelings about it either way.

jensengar: Alright. Go ahead a leave this then.

src/notifications.js
```diff
@@ -2,6 +2,7 @@ import { requestApi } from './utils';
 const debounce = require('debounce-promise');
 
 const PUBSUB_HOST_DEFAULT = 'notifications.mypurecloud.com';
+const MAX_SUBSCRIBABLE_TOPICS = '1000';
```
djhouseknecht, line 5: Is there a reason this is a string?

jensengar: Good catch

caitlanspronk: Nope. Mistake that I just corrected.

src/notifications.js
JackieChiles: Right :+1: It's just to make sure that we don't have a "dropped topics" message that's too big to send.

PR #54, opened 2020-04-13
Engageui 3797 topic prioritization

caitlanspronk:


jensengar:
So this is a bit different than I had imagined, but I can see why it ended up this way. I think the SOP for creating a prioritized topic should happen when you actually subscribe, e.g. `subscribe: function (topic, handler, immediate = false, priority = 0) {...`. Is there a case for *changing* the priority of a topic after the initial subscription?

src/notifications.js
```diff
@@ -279,6 +328,29 @@ export default class Notification {
...
+                oldPriorities[postfix] = newPriority;
+              }
+            });
+          } else {
+            const newTopics = {};
+            topicParts.postfixes.forEach(p => {
+              newTopics[p] = newPriority;
+            });
+            this.topicPriorities[topicParts.prefix] = newTopics;
+          }
```
jensengar, line 352: Do we need to resubscribe here? My thinking is if you added priority to a topic that wasn't high enough to make the cut list (more than 1k subscriptions), then I would think we would want to resubscribe to the new list of topics.

caitlanspronk:
@jensengar I see what you're saying. I can add the priorities as parameter for both `subscribe` and `bulksubscribe` and not expose `setTopicPriorities`. I don't think we need to give consumers the ability to change a topic's priority after initial subscription.

src/notifications.js
```diff
@@ -222,9 +253,48 @@ export default class Notification {
...
+        }, {});
+        this.topicPriorities[topicParts.prefix] = newTopics;
+      }
+    });
+  }
+
   get expose () {
     return {
-      subscribe: function (topic, handler, immediate) {
+      subscribe: function (topic, handler, immediate, priority) {
```
jensengar, line 295: will you add a test for this new param please?

PR #58, opened 2020-06-15
[ENGAGEUI-4023] unsubscribe from precombined topics

caitlanspronk:


test/unit/notifications.test.js
```diff
@@ -418,6 +418,28 @@ test('notifications | createSubscription should correctly register handlers for
   t.is(notification.subscriptions['v2.users.8b67e4d1-9758-4285-8c45-b49fedff3f99'], undefined);
 });
```
jensengar, line 420: Is there a use case for unsubscribing to a single topic in a combined topic? Example: unsubscribing to `geolocation` in `v2.users.731c4a20-e6c2-443a-b361-39bcb9e087b7?geolocation&presence&routingStatus&conversationsummary` but keeping the rest. Seems like that would be valid to me. If so, can you create a test for that please?

caitlanspronk: I missed this comment, sorry. I can't think of why that wouldn't be a valid use case, so I added a test for that.

PR #59, opened 2020-06-29
Log http requests

AndrewGnagy:
This adds middlewhere for superagent that will log out http requests, such as those used for the bulk subscription method.

### DIFF
--- a/src/request-logger.js
+++ b/src/request-logger.js
@@ -11,7 +11,7 @@ function attachSuperagentLogger (logger, req) {
     timestamp,
     method.toUpperCase(),
     req.url,
-    JSON.stringify(req.body));
+    JSON.stringify(req.toJSON().data));
 
   req.on('response', function (res) {
     let now = new Date().getTime();
--- a/src/request-logger.js
+++ b/src/request-logger.js
@@ -15,7 +15,7 @@ function attachSuperagentLogger (logger, data, req) {
     data);
 
   req.on('response', function (res) {
-    let now = new Date().getTime();
+    let now = new Date().toISOString();
     let elapsed = now - start;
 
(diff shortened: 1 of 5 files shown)


src/request-logger.js
```diff
@@ -0,0 +1,31 @@
...
+  data = data || {};
+  return attachSuperagentLogger.bind(null, logger, data);
+};
+
+function attachSuperagentLogger (logger, data, req) {
+  let start = new Date().getTime();
+  let timestamp = new Date().toISOString();
+  let method = req.method;
+
+  logger.info('%s >>> %s %s \n Body: %s',
```
jensengar, line 12: So the logger that runs inside of streaming client doesn't support lots of args. The method signature of that logger looks like this: https://github.com/purecloudlabs/genesys-cloud-client-logger/blob/master/src/logger.ts#L72

We also have a lot of inconsistencies with how we log things and it causes issue. We try and log a sumo-searchable string as the message then put the details as a json object:
```
      this.logger.info('Streaming client disconnected.', { channelId });
```
 
So in this case, I would do something like:
```
logger.info(`request: ${method.toUppercase} ${req.url}`, { timestamp, data });
```

Also, would these logs fit into `debug` level? If not, we should make the request logging an opt-in per request thing.

src/request-logger.js
```diff
@@ -0,0 +1,31 @@
...
+    timestamp,
+    method.toUpperCase(),
+    req.url,
+    data);
+
+  req.on('response', function (res) {
+    let now = new Date().toISOString();
+    let elapsed = now - start;
+
+    logger.info('%s <<< %s %s %s %s \n Correlation-id: %s \n Body: %s',
```
jensengar, line 22: same as above

PR #60, opened 2020-07-01
PCM-1205

djhouseknecht:


babel.config.js
```diff
@@ -0,0 +1,27 @@
...
+            'IE 11',
+            'not dead'
+          ]
+        }
+      ]
+    ],
+    plugins: [
+      ['@babel/plugin-transform-runtime', {
+        /* if we are testing, we don't want core-js polyfills */
+        corejs: api.env() === 'test' ? false : 3
```
djhouseknecht, line 23: This will exclude polyfills when we are testing.

webpack.config.js
```diff
@@ -4,18 +4,39 @@ const WebpackAutoInject = require('webpack-auto-inject-version');
...
+      this is so babel doesn't try to polyfill/transpile core-js (which is the polyfill)
+        and the build tools.
+      But we want it polyfill/transpile all other node_modules when building for the web
+    */
+    babelExcludes = [
+      /\bcore-js\b/,
+      /\bwebpack\/buildin\b/
+    ];
+
+    filename += '.bundle';
```
djhouseknecht, line 24: I went back and forth on the naming for the cdn build. It was either `streaming-client.bundle.js` or `streaming-client.cdn.js`. Bundle sounded better to me, but I am open to opinions.

package.json
```diff
@@ -1,14 +1,13 @@
...
+  "version": "12.0.0",
   "description": "client for the PureCloud Streaming APIs (websocket/xmpp interface)",
   "repository": "https:github.com/purecloudlabs/purecloud-streaming-client",
   "license": "MIT",
   "publishConfig": {
     "registry": "https://registry.npmjs.org/"
   },
-  "main": "dist/streaming-client.min.cjs",
-  "browser": "dist/streaming-client.js",
+  "main": "dist/src/client.js",
```
djhouseknecht, line 10: I decided to just point at the es6 module and let the consuming application transpile as it needs.

webpack.config.js
```diff
@@ -4,18 +4,39 @@ const WebpackAutoInject = require('webpack-auto-inject-version');
...
+  } else {
+    /* if we are building for 'module', don't polyfill/transpile any dependencies */
+    babelExcludes = [/node_modules/];
+  }
+
+  filename += minimize ? '.min.js' : '.js';
+
   return {
-    target: node ? 'node' : 'web',
+    target: cdn ? 'web' : 'node',
```
djhouseknecht, line 33: I am not 100% sure if we want/need to build a node version. Open to opinions. We could just bundle for the cdn and let any consumers use the src.

webpack.config.js
jensengar: yeah, i think bundle is better.

PR #64, opened 2020-08-03
PCM-1390

djhouseknecht:
Lessons learned from "building to node"

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,13 +1,13 @@
 {
   "name": "genesys-cloud-streaming-client",
-  "version": "12.0.0",
+  "version": "12.0.1",
   "description": "client for the Genesys Cloud Streaming APIs (websocket/xmpp interface)",
   "repository": "https:github.com/purecloudlabs/genesys-cloud-streaming-client",
   "license": "MIT",
   "publishConfig": {
     "registry": "https://registry.npmjs.org/"
   },
-  "main": "src/client.js",
+  "main": "dist/streaming-client.js",
   "files": [
     "src",
     "dist"
@@ -49,7 +49,7 @@
     "backoff-web": "^1.0.1",
     "core-js": "^3.6.5",
     "debounce-promise": "^3.1.2",
-    "genesys-cloud-streaming-client-webrtc-sessions": "^8.0.0",
+    "genesys-cloud-streaming-client-webrtc-sessions": "^8.0.1",
     "limiter": "^1.1.0",
     "stanza": "^11.0.0",
     "superagent": "^5.1.0",
@@ -75,7 +75,8 @@
     "semistandard": "^13.0.1",
     "webpack": "^4.35.3",
     "webpack-auto-inject-version": "^1.2.2",
-    "webpack-cli": "^3.3.5"
+    "webpack-cli": "^3.3.5",
+    "webpack-node-externals": "^2.5.0"
   },
   "false": {}
 }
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,13 +1,13 @@
...
-  "version": "12.0.0",
+  "version": "12.0.1",
   "description": "client for the Genesys Cloud Streaming APIs (websocket/xmpp interface)",
   "repository": "https:github.com/purecloudlabs/genesys-cloud-streaming-client",
   "license": "MIT",
   "publishConfig": {
     "registry": "https://registry.npmjs.org/"
   },
-  "main": "src/client.js",
+  "main": "dist/streaming-client.js",
```
djhouseknecht, line 10: Had to change this to point at the node compatible, webpack built source file. Pointing straight at `src/client.js` caused a lot of headaches for build tools trying to load it in a node way (using commonjs).

webpack.config.js
```diff
@@ -21,21 +27,59 @@ module.exports = (env) => {
...
+      presets: ['@babel/preset-env']
+    };
   }
 
   filename += minimize ? '.min.js' : '.js';
+  mode = minimize ? 'production' : 'development';
 
   return {
-    target: cdn ? 'web' : 'node',
+    target: 'web',
```
djhouseknecht, line 76: After a ton of trial and error, this was one of the main pieces I was missing. Even though we are building to a "bundled/cdn" version _and_ a "node/commonjs" version, we still need to target `'web'`. This will allow webpack to use 3rd party libs' `package.json#browser` field. It causes a ton of build headaches for consumers of the streaming-client if the target is `'node'` (because the builders can't find `crypto`, `node-stringprep`, and a few others).

webpack.config.js
```diff
@@ -21,21 +27,59 @@ module.exports = (env) => {
...
   } else {
-    /* if we are building for 'module', don't polyfill/transpile any dependencies */
-    babelExcludes = [/node_modules/];
+    /* we are building for node */
+    const modulesToBundle = [/* left this here just in case it is needed */];
+
+    /* we don't want to bundle node_modules */
+    externals.push(nodeExternals({
+      allowlist: modulesToBundle.map(m => new RegExp(`^${m.replace(/-/g, '\\-')}`))
+    }));
```
djhouseknecht, line 59: The `externals` along with [webpack-node-externals](https://www.npmjs.com/package/webpack-node-externals) was the main missing piece I couldn't figure out. This tells webpack to not bundle/build dependencies. This allows consumers to build and transpile the streaming-client using any node build tool they like (like lasso, webpack. rollup, ember, etc). 

The other missing was the `target: 'web'`. [See my other comment below](https://github.com/purecloudlabs/genesys-cloud-streaming-client/pull/64/files#r464675419)

webpack.config.js
```diff
@@ -21,21 +27,59 @@ module.exports = (env) => {
...
+            ]
+          }
+        ]
+      ],
+      plugins: [
+        ['@babel/plugin-transform-runtime', {
+          corejs: 3
+        }]
+      ]
+    };
```
djhouseknecht, line 51: We want all these polyfils (core-js) and babel's runtime if we are deploying to the cdn.

webpack.config.js
```diff
@@ -21,21 +27,59 @@ module.exports = (env) => {
...
+
+    /* if we are building for 'module', don't polyfill/transpile most dependencies */
+    babelExcludes = [
+      new RegExp(`/node_modules/(?!${modulesToBundle.join('|')}/)`)
+    ];
+
+    babelOptions = {
+      sourceType: 'unambiguous',
+      presets: ['@babel/preset-env']
+    };
```
djhouseknecht, line 69: If we aren't deploying to the cdn, we just want the bare minimum babel config.

PR #66, opened 2020-08-19
[PCM-1388] hack around the angular imports issue

jensengar:


src/stanzaio-light.js
```diff
@@ -9,6 +9,17 @@ import { Client } from 'stanza/browser-module';
...
+// HACK: for some reason, in production builds of angular, the imports for these plugins are getting messed.
+// A lot of time has been sunk into figuring out why/how to fix it properly, but to no avail. This is a
+// harmless hack that works around the issue.
+export function getActualFunction (dep) {
+  if (dep.default) {
+    return dep.default;
+  }
+
+  return dep;
+}
```
djhouseknecht, line 21: Do we need to move the `.bind()` call into here at all? I am assuming no since I saw it work earlier.

jensengar: No. the bind was because it was the function being passed to the `map`, instead of an inline fn.

PR #69, opened 2020-10-06
Stanza12

jensengar:


djhouseknecht (approved):
I left some comments, but they probably aren't blocking anything from merging. This all looks good and I can tell took some work!

src/client.ts
```diff
@@ -27,10 +29,12 @@ function stanzaioOptions (config) {
     jid: config.jid,
     credentials: {
       username: config.jid,
-      password: `authKey:${config.authToken}`
+      // TODO: don't commit this
+      password: `authKey:${config.authToken}asdfds`
```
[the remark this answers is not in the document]
jensengar: ty. I put this in to try and make auth fail... but it turns out hawk doesn't care about the password and just relies on the channelId which is created earlier with an api call.

src/client.ts
```diff
@@ -179,7 +196,7 @@ export default class Client {
         return;
```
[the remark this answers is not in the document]
jensengar: I don't know what you mean here

src/media-session.ts
```diff
@@ -0,0 +1,90 @@
...
+        info: {
+          infoType: JINGLE_INFO_ACTIVE
+        }
+      });
+    }
+
+    super.onIceStateChange();
+  }
+
+  onIceCandidate (e: RTCPeerConnectionIceEvent) {
```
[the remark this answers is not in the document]
jensengar: stats gatherer does that.

src/media-session.ts
```diff
@@ -0,0 +1,90 @@
...
+        // find a transceiver whose receiver is the same kind but sender doesn't have a track
+        const transceiver = this.pc.getTransceivers().find(transceiver => !transceiver.sender.track && transceiver.receiver.track?.kind === track.kind);
+        sender = transceiver?.sender;
+      }
+
+      if (sender) {
+        return sender.replaceTrack(track);
+      }
+
+      await (this.pc as any).addTrack(track, stream);
```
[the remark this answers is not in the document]
jensengar: This is my bad. There are some native definitions for the peerConnection that are wrong and I assumed this was one of them.

src/utils.ts
```diff
@@ -51,6 +51,29 @@ export function splitIntoIndividualTopics (topicString) {
...
+};
+
+export const isPeerVideoJid = function (jid: string): boolean {
+  return isVideoJid(jid) && jid.startsWith('peer-');
+};
+
+export const isVideoJid = function (jid: string): boolean {
+  return !!(jid && jid.match(/@conference/) && !isAcdJid(jid));
+};
+
```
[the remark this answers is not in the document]
jensengar: agreed.

src/client.ts
```diff
@@ -103,58 +122,52 @@ export default class Client {
...
     this.on('_disconnected', (event) => {
+      if (this._stanzaio.transport || this.connecting) {
+        this.logger.info('disconnected event received, but reconnection is in progress');
+        return;
+      }
+
       this.connected = false;
       this._ping.stop();
 
       // example url: "wss://streaming.inindca.com/stream/channels/streaming-cgr4iprj4e8038aluvgmdn74fr"
```
kglawrence, line 133: Is this comment still relevant here?

kglawrence (approved):

djhouseknecht (approved):

src/types/media-session.ts
```diff
@@ -103,3 +85,19 @@ export class GenesysCloudMediaSession extends MediaSession {
     });
   }
 }
+
+export interface SessionEvents {
```
jensengar, line 89: how this works is `[eventName]: parameters`. there's a little weirdness here with they syntax. basically, if there's only one param, you can just put that there, e.x.: `mute: JingleInfo` means you can do this:

```
session.on('mute', (info: JingleInfo) => { ... });
```
but won't allow 
```
session.on('mute', (info: JingleInfo, somethingElse: string) => { ... });
```

if you specific a function in the definition, it allows you to specify multiple params. ex:
```
peerTrackAdded: (track: MediaStreamTrack, stream?: MediaStream) => void;
```
allows you to do 
```
session.on('peerTrackAdded', (track, stream) => { ... })
```

src/types/media-session.ts
```diff
@@ -103,3 +85,19 @@ export class GenesysCloudMediaSession extends MediaSession {
...
+  unmute: JingleInfo;
+  sessionState: 'starting' | 'pending' | 'active';
+  connectionState: 'starting' | 'connecting' | 'connected' | 'interrupted' | 'disconnected' | 'failed';
+  terminated: JingleReason;
+  stats: StatsEvent;
+  endOfCandidates: void;
+}
+
+applyMixins(GenesysCloudMediaSession, [ EventEmitter ]);
+export interface GenesysCloudMediaSession extends StrictEventEmitter<EventEmitter, SessionEvents> { }
```
jensengar, line 103: These two lines are derived from the typescript cookbook. basically it is a way to do multiple inheritance. Since we are already extending jingle's `MediaSession` and we *also* was to extend/add an `EventEmitter` we basically specify it in the interface and then do a javascripty way of extending the class at runtime.

src/types/media-session.ts
djhouseknecht: That is an awesome find! I've been wanting to find a way to strongly type callback functions/events like this. 🧨

src/types/media-session.ts
```diff
@@ -0,0 +1,103 @@
...
+  setupStatsGatherer () {
+    this.statsGatherer = new StatsGatherer(this.pc);
+    this.statsGatherer.on('stats', this.emit.bind(this, 'stats'));
+  }
+
+  onIceStateChange () {
+    const state = this.pc.iceConnectionState;
+
+    if (state === 'connected') {
+      this._log('info', 'sending session-info: active');
```
djhouseknecht, line 31: If this log is getting sent to Sumo, is there any identifying info we can send with it like a sessionId or convoId?

src/webrtc.ts
```diff
@@ -0,0 +1,482 @@
...
+
+    this.client._stanzaio.on('jingle:outgoing', session => {
+      return this.emit(events.OUTGOING_RTCSESSION, session);
+    });
+
+    this.client._stanzaio.on('jingle:incoming', session => {
+      return this.emit(events.INCOMING_RTCSESSION, session);
+    });
+
+    const eventsToProxy = [
```
djhouseknecht, line 184: Since we are using a strict event emitter, it may be nice to have these events strongly typed. I think something like this will work: 
``` ts
const eventsToProxy: Array<keyof SessionEvents> = [ ... ];
```

djhouseknecht (approved):

src/types/media-session.ts
jensengar: we'll need to go through all the logs in streaming client. I'd like to also add the genesys-cloud-client-logger to streaming client as soon as we get this stanza stuff straightened out.

PR #70, opened 2020-10-16
add webpack build and needed deps to generate an ie compat build

djhouseknecht:


package.json
```diff
@@ -80,17 +79,20 @@
     "lru-cache": "^6.0.0",
     "stanza": "^12.13.1",
     "superagent": "^5.1.0",
+    "unorm": "^1.6.0",
```
djhouseknecht, line 82: [unorm](https://github.com/walling/unorm/) is a little outdated, but we only need it to make normalization work in IE which is also outdated.

jensengar:
As I think about this, we may have to change the `browser` field in the package.json to point to this new build file. Genesys Cloud is going to need to use the IE safe version also. Volt can use the es6 version though.

jensengar:
I'm not sure which field ember looks at though, we may need to do something different, but that can wait until we start working it into epc.

jensengar (approved):

djhouseknecht:
I _think_ ember will bundle all the deps to be IE compatible regardless of how they are built. But you are right, we can cross that bridge when we get there.

PR #72, opened 2020-10-21
Stats gatherer2

jensengar:


package.json
```diff
@@ -71,10 +71,10 @@
   "author": "",
   "dependencies": {
     "@babel/runtime-corejs3": "^7.10.4",
+    "async": "^3.2.0",
```
djhouseknecht, line 74: I asked in chat already, but where is this used? I don't see any imports in the source code, but I know that doesn't mean it isn't used.

djhouseknecht (approved):

kglawrence (approved):

PR #77, opened 2020-12-18
[PCM-1498] fix ice transport policy

jensengar:


src/webrtc.ts
```diff
@@ -89,13 +89,13 @@ export class WebrtcExtension extends EventEmitter {
 
   configureStanzaJingle () {
     Object.assign(this.client._stanzaio.jingle.config.peerConnectionConfig, {
-      sdpSemantics: 'unified-plan',
-      iceTransportPolicy: this.config.iceTransportPolicy || 'all'
```
jensengar, line 93: the problem here is iceTransportPolicy isn't set at bootstrap and this is the only place this is set.

ckbrewer33 (approved):

src/webrtc.ts
```diff
@@ -89,13 +89,13 @@ export class WebrtcExtension extends EventEmitter {
...
+      sdpSemantics: 'unified-plan'
     });
   }
 
   prepareSession (options: any) {
-    options.iceServers = this.config.iceServers || options.iceServers;
+    options.config.iceServers = this.config.iceServers || options.iceServers;
+    options.config.iceTransportPolicy = this.config.iceTransportPolicy || 'all';
```
jensengar, line 98: update the iceTransportPolicy for every new session.

PR #79, opened 2021-01-12
[PCM-1502] fix build

jensengar:


djhouseknecht (approved):
This looks good to me. With the `"main"` change, I can pull it into via node. We won't really know if all the dependencies are node compatible until we actually try to use the streaming-client in a node environment.

PR #80, opened 2021-01-20
[PCM-1510] pass jid resource to stanza

jensengar:


src/client.ts
```diff
@@ -20,10 +20,11 @@ let extensions = {
   webrtcSessions: WebrtcExtension
 };
 
-function stanzaioOptions (config) {
+function stanzaioOptions (config: ClientOptions & { channelId: string }): AgentConfig {
   let wsHost = config.host.replace(/\/$/, '');
-  let stanzaOptions = {
+  let stanzaOptions: AgentConfig = {
     jid: config.jid,
+    resource: config.jidResource,
```
djhouseknecht, line 27: The fact that this is already spec'ed out is awesome. Great find, John Hjelle.

src/client.ts
```diff
@@ -80,6 +81,7 @@ export interface ClientOptions {
   authToken?: string;
   jwt?: string;
   jid?: string;
+  jidResource?: string;
```
djhouseknecht, line 84: Would it be too much to add code comments on these options? I find that really helpful for the code hinting and if I need to `cmd + click` the definition to take me to the `*.d.ts` file. 

Also wouldn't hurt to type out the `config` for the `Client` class.

djhouseknecht (changes requested):
Two overal comments: 
1. I think the CHANGELOG needs to be updated with these changes
2. Does the version need to be bumped (since we don't get that for free anymore)?

src/client.ts
jensengar: Someday we'll have to change the doc to match whatever you are doing with the sdk.

djhouseknecht (approved):

PR #81, opened 2021-02-05
PCM-1516 fixing hard reconnects

djhouseknecht:


src/ping.ts
```diff
@@ -36,7 +36,7 @@ export class Ping {
       };
       this.client.logger.warn('Missed a ping.', Object.assign({ error: err }, info));
       if (++this.numberOfFailedPings > this.failedPingsBeforeDisconnect) {
-        clearInterval(this.pingIntervalId);
+        this.stop();
```
djhouseknecht, line 39: `this.stop()` clears the interval and sets it to `null` which I thought was more inclusive.

src/reconnector.ts
```diff
@@ -107,11 +125,54 @@ export class Reconnector {
...
+
+    this.cleanupReconnect();
     this._hasConnected = false;
-    return this.client.connect();
+
+    this._hardReconnectRetryInfo = {} as any;
+    this._hardReconnectRetryInfo!.promise = new Promise<void>((resolve, reject) => {
+      /* defer this to allow it to be canceled */
+      this._hardReconnectRetryInfo!.resolve = resolve;
+      this._hardReconnectRetryInfo!.reject = reject;
```
djhouseknecht, line 140: I didn't want to defer this promise, but if we clear the interval without the promise ever completing, then calls to `hardReconnect()` could end up pending indefinitely. A pretty common scenario is if the hard reconnect interval is running, and then someone calls `client.disconnect()`. We would want to stop any pending `hardReconnect` calls. 

On the fence about whether to `resolve()` or `reject()` if someone cancels a pending `hardReconnect()`. Thoughts?

src/reconnector.ts
```diff
@@ -128,7 +189,20 @@ export class Reconnector {
   }
 
   stop () {
-    this.backoff.reset();
-    this._backoffActive = false;
+    this.cleanupReconnect();
+    this._stopHardReconnect();
+  }
+
+  private _stopHardReconnect (error?: any) {
```
djhouseknecht, line 196: see note above regarding the deferred promise here.

src/reconnector.ts
```diff
@@ -18,11 +21,29 @@ export const CXFRDefinition: DefinitionOptions = {
 
 export class Reconnector {
   client: Client;
-  backoff: any;
+  backoff: Backoff;
```
jensengar, line 24: 🔥

src/reconnector.ts
```diff
@@ -107,11 +125,54 @@ export class Reconnector {
...
+
+          this._stopHardReconnect();
+        } catch (error) {
+          /* this error comes from superagent for requests that timeout for network offline */
+          if (error.message.startsWith('Request has been terminated')) {
+            return this.client.logger.debug('request offline. attempting reconnect again', error);
+          }
+          /* error client thrown timeouts */
+          else if (error.message.startsWith('Timeout: ')) {
+            return this.client.logger.debug(`Streaming-client timedout. attempting reconnect again: "${error.message}"`);
```
jensengar, line 162: NIT: timed out

src/reconnector.ts
jensengar: If someone calls disconnect, we should reject with some sort of `Cancelled` error.

djhouseknecht: 👍  will do

index.html
```diff
@@ -0,0 +1,12 @@
+<html>
```
djhouseknecht, line 1: super helpful test html files for quick local debugging (this redirects to `test/index.html`)

djhouseknecht:
I ran `npm audit fix` which cleaned a lot in package-lock.json

package.json
```diff
@@ -78,7 +80,7 @@
     "lru-cache": "^6.0.0",
     "stanza": "^12.13.1",
     "strict-event-emitter-types": "^2.0.0",
-    "superagent": "^5.1.0",
+    "superagent": "^6.1.0",
```
djhouseknecht, line 83: bumped this while I was here

jensengar (approved):

PR #82, opened 2021-02-09
[PCM-1525] fix conversationId param on upgrademedia stanza

jensengar:


djhouseknecht (changes requested):
Looks good. Just need to update the CHANGELOG.

djhouseknecht (approved):

PR #83, opened 2021-02-10
PCM-1527 PCM-1516 …

djhouseknecht:
added retry logic to http requests. re-wrote hard reconnect logic to reuse http retry logic


package.json
```diff
@@ -26,11 +26,11 @@
...
     "compile:webpack": "webpack --mode production",
     "compile:webpack:ie": "webpack --env.ie --mode production",
-    "test": "semistandard && npm run test:unit",
+    "test": "npm run lint && npm run test:unit",
     "test:unit": "jest --runInBand",
     "test:watch": "jest --watch --runInBand --collectCoverage=false",
-    "lint": "semistandard",
-    "lint:fix": "semistandard --fix",
+    "lint": "tslint --project . --config tslint.json",
+    "lint:fix": "npm run lint -- --fix",
```
djhouseknecht, line 33: Changed this to tslint so it actually lints our source code.

src/client.ts
```diff
@@ -101,6 +100,7 @@ export class Client {
   config: any;
   streamId: any;
 
+  http: HttpClient;
```
djhouseknecht, line 103: Moved all http request logic into a separate class so retry state can be manage easier.

src/http-client.ts
```diff
@@ -0,0 +1,78 @@
...
+    503,
+    504
+  ]);
+
+  async requestApiWithRetry (path: string, opts: RequestApiOptions, retryInterval?: number): Promise<any> {
+    const request = retryPromise<any>(
+      this.requestApi.bind(this, path, opts),
+      (error: any) => error && HttpClient.retryStatusCodes.has(error.status),
+      retryInterval
+    );
```
djhouseknecht, line 34: See `retryPromise()` for the actual retry interval logic. Essentially, we want to keep retrying `this.requestApi()` until it resolves or rejects with an error other than a retriable status code.

src/http-client.ts
```diff
@@ -0,0 +1,78 @@
...
+  data?: any;
+  host: string;
+  version?: string;
+  contentType?: string;
+  authToken?: string;
+  logger?: any
+};
+
+export class HttpClient {
+  private _apiRequestsMap = new Map<string, RetryPromise<any>>();
```
djhouseknecht, line 17: This map holds all promises that are in a retry state. Once a promise resolves/rejects, it is removed from the map. This is here so any pending retries can be stopped later.

src/reconnector.ts
```diff
@@ -26,23 +29,7 @@ export class Reconnector {
...
-  _retryStatusCodes = new Set([
-    408,
-    413,
-    429,
-    500,
-    502,
-    503,
-    504,
-  ]);
+  _hardReconnectRetryInfo: null | RetryPromise<void> = null;
```
djhouseknecht, line 32: Moved all this logic to the `utils.retryPromise` utility function.

src/reconnector.ts
```diff
@@ -163,52 +149,53 @@ export class Reconnector {
...
-      /* superagent retriable error */
-      else if (error && this._retryStatusCodes.has(error.status)) {
-        return this.client.logger.debug('Received HTTP status code eligible for retry. attempting reconnect again', error);
-      }
+    await this.client.connect();
+  }
 
-      /* if it is an error we can't retry, reject with it */
-      this._stopHardReconnect(error);
+  private _shouldRetryError (error: Error): boolean {
```
djhouseknecht, line 187: If this returns `true`, it means we need to keep trying the hard reconnect. Main use cases: 
1. network goes offline
2. 429s

src/utils.ts
```diff
@@ -81,6 +52,54 @@ export const isVideoJid = function (jid: string): boolean {
...
+  cancel: (reason?: string | Error) => void;
+  complete: (value?: T) => void;
+  _id: string;
+};
+
+export function retryPromise<T> (
+  promiseFn: () => Promise<T>,
+  retryFn: (error?: Error | any) => boolean,
+  retryInterval: number = 15000
+): RetryPromise<T> {
```
djhouseknecht, line 66: Main gist of `retryPromise`: 
1. Call the first param (which is a function that returns a `Promise`) until it resolves/rejects
2. Second param is a callback that is passed any errors received from the promise callback. If the call back returns `true`, it will retry the promise function
3. Retry on the interval passed in

It returns a promise that resolves/rejects when the promise callback does _OR_ if the returned `cancel() or complete()` functions are called which will short circuit the retry logic.

src/ping.ts
```diff
@@ -34,7 +36,13 @@ export class Ping {
         channelId: this.client.config.channelId,
         jid: this.client._stanzaio.jid
       };
-      this.client.logger.warn('Missed a ping.', Object.assign({ error: err }, info));
+
+      /* need to check this because we could have already stopped the pinging. In which case, we don't want to count the missed ping */
+      if (intervalIdAtStart !== this.pingIntervalId) {
+        return this.client.logger.warn('Disregarding a missed ping for an old interval', Object.assign({ error: err }, info));
+      }
```
djhouseknecht, line 43: I needed to add this for pings that stanza would reject _after_ we had already called `ping.stop()`. This will drop it on the floor in that case.

ckbrewer33 (approved):

src/client.ts
```diff
@@ -133,7 +137,7 @@ export class Client {
       channelId: null // created on connect
     };
 
-    this.on('disconnected', (event) => {
```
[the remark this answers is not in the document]
djhouseknecht: It is `undefined` so no real use.

src/http-client.ts
```diff
@@ -0,0 +1,78 @@
+import request from 'superagent';
+
+import reqlogger from './request-logger';
+import { RetryPromise, retryPromise } from './utils';
+
+export type RequestApiOptions = {
+  method?: string;
```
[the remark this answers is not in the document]
djhouseknecht: Will do 👍

src/http-client.ts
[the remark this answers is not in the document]
djhouseknecht: 👍

PR #84, opened 2021-02-26
PCM-1540 convert http errors into more readable error objects

djhouseknecht:


src/http-client.ts
```diff
@@ -40,14 +47,75 @@ export class HttpClient {
       .set('Authorization', `Bearer ${opts.authToken}`)
       .type(opts.contentType || 'json');
 
-    return response.send(opts.data); // trigger request
+    return response.send(opts.data) // trigger request
+      .catch(err => { throw this.formatRequestError(err); });
```
djhouseknecht, line 51: This will format the errors in the console so they are more useful.

src/types/interfaces.ts
```diff
@@ -7,3 +7,51 @@ export type RequestApiOptions = {
...
+  responseBody: string;
+  method: string;
+  requestBody: string;
+  url: string;
+}
+
+export interface IError {
+  message: string;
+  name: string;
+  stack?: string;
```
djhouseknecht, line 56: If the caller sends these errors straight up to sumo, it gets pretty ugly. But I didn't want to just remove the stack altogether from the returned object.

src/http-client.ts
```diff
@@ -40,14 +47,75 @@ export class HttpClient {
...
+
+        url: res.error.url,
+        message: res.error.message,
+        name: res.error.name,
+        stack: res.error.stack
+      };
+    }
+
+    /* if we don't have a superagent error */
+    return {
```
jensengar, line 93: is there a reason to not just return `Error` here?

djhouseknecht: I was torn here. My thought was: this function wants to format all errors, not just `superagent` errors. By formatting the error this way, it will hit the console a lot cleaner.

djhouseknecht: Albeit, this does mess up the prototype chain making `error instanceof Error` return `false`. I can switch it back if you think that would be helpful

jensengar: I think we ought to just leave it as an error, but I don't have particularly strong feelings either way.

jensengar (approved):

PR #85, opened 2021-03-01
[PCM-1518] - Optimize sending of webrtc stats.

zservies:


src/webrtc.ts
```diff
@@ -98,13 +107,26 @@ export class WebrtcExtension extends EventEmitter {
...
+    const session = new GenesysCloudMediaSession(
+      options,
+      this.getSessionTypeByJid(options.peerID),
+      this.config.allowIPv6
+    );
     this.proxyStatsForSession(session);
     return session;
   }
 
+  calculatePayloadSize (trace: any): number {
```
jensengar, line 122: NitPick: I don't think this really belongs as a class fn. It's probably better suited as an exported util. I'd put it in `utils.ts`.

src/webrtc.ts
```diff
@@ -146,8 +197,33 @@ export class WebrtcExtension extends EventEmitter {
         logger: this.client.logger,
         data
       });
+      if (this.currentMaxStatSize === desiredMaxStatsSize) {
```
jensengar, line 200: This testing code needs to be removed.

src/webrtc.ts
```diff
@@ -197,9 +273,12 @@ export class WebrtcExtension extends EventEmitter {
       'endOfCandidates'
     ];
     for (const e of eventsToProxy) {
-      this.client._stanzaio.jingle.on(e, (session: GenesysCloudMediaSession, ...data: any) => {
-        session.emit(e as any, ...data);
-      });
+      this.client._stanzaio.jingle.on(
```
jensengar, line 276: Can you change your prettier plugin or what ever this is so it doesn't affect "unchanged" code? Otherwise this will probably get pretty messing with bigger projects.

test/unit/webrtc.test.ts
```diff
@@ -757,9 +757,44 @@ describe('proxyStatsForSession', () => {
       actionName: 'test'
     });
 
-    expect(webrtc['statsToSend']).toEqual([formattedStats]);
+    expect(webrtc['statsArr']).toEqual([formattedStats]);
     expect(webrtc['throttledSendStats']).toHaveBeenCalled();
   });
+
+  it('should flush throttledSendStats', async () => {
```
jensengar, line 764: nice job here.

test/unit/webrtc.test.ts
```diff
@@ -769,38 +804,51 @@ describe('sendStats', () => {
     const webrtc = new WebrtcExtension(client as any);
 
     const sendSpy = jest.spyOn(client.http, 'requestApi').mockResolvedValue(null);
-
-    webrtc['statsToSend'].push({} as any);
+    webrtc['statsArr'].push({} as any);
     webrtc['throttledSendStats']();
     expect(sendSpy).not.toHaveBeenCalled();
-    await wait(3050);
+    await wait(25050);
```
jensengar, line 810: we should probably modify this. I think there's value is actually testing against a real timer, but adding a 25 sec delay in a test is pretty gnarly. Here, we probably want to use jest's fake timers. I think there's also a restoreTimers fn or something equivalent, so I'd start the test with fake timers, then restore at the end of the test.

jensengar (changes requested):

test/unit/webrtc.test.ts
djhouseknecht: There are some example usages of fake timers in `test/unit/ping.test.ts` and `test/unit/reconnector.test.ts`. Just be careful if you start mixing them with promises. Things can get messy and confusing. I've spent a couple days working through fake timers and promise issues in tests.

djhouseknecht:
When you merge this, can you squash the commits and leave a single commit message? It will help keep the master branch history a little more clean. 

image

src/webrtc.ts
zservies: Yeah I'm going to just disable it for now, not sure what the deal is with its config right now.

src/webrtc.ts
zservies: Moved it.

djhouseknecht (approved):

test/unit/webrtc.test.ts
zservies: In previous versions of Jest there was some weird behavior with mock timers and debounce/throttle but version 26+ seems to have resolved it. I've updated the tests with mock timers.

jensengar (approved):

PR #87, opened 2021-03-09
PCM-1540 

djhouseknecht:
added noAuthHeader param to http reqeusts. exported http-client and parseJwt from index.ts

### DIFF
--- a/src/http-client.ts
+++ b/src/http-client.ts
@@ -43,8 +43,12 @@ export class HttpClient {
   requestApi (path: string, opts: RequestApiOptions): Promise<any> {
     let response = request[opts.method](this._buildUri(opts.host, path, opts.version))
       .use(reqlogger.bind(this, opts.logger, opts.data))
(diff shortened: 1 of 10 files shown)


src/http-client.ts
```diff
@@ -43,9 +43,13 @@ export class HttpClient {
   requestApi (path: string, opts: RequestApiOptions): Promise<any> {
     let response = request[opts.method](this._buildUri(opts.host, path, opts.version))
       .use(reqlogger.bind(this, opts.logger, opts.data))
-      .set('Authorization', `Bearer ${opts.authToken}`)
       .type(opts.contentType || 'json');
 
+    // default to include auth header
+    if (!opts.noAuthHeader) {
+      response.set('Authorization', `Bearer ${opts.authToken}`);
+    }
```
djhouseknecht, line 51: I'm going to be using this http-client in the sdk and the sdk needs to be able to make requests without an authorization header.

jensengar (approved):

PR #88, opened 2021-03-10
PCM-1552 reworked ping logic to avoid chrome 88 timer throttling

djhouseknecht:


test/unit/ping.test.ts
```diff
@@ -44,232 +41,201 @@ describe('Ping', () => {
...
-    jest.advanceTimersByTime(PING_INTERVAL_WITH_BUFFER);
-    expect(pingCallCount).toBe(1);
-    ping.start();
-    jest.advanceTimersByTime(PING_INTERVAL_WITH_BUFFER);
-    expect(pingCallCount).toBe(2);
+  describe('constructor()', () => {
+    it('accepts null options', () => {
+      const ping = new Ping({} as any);
+      expect(ping).toBeTruthy();
+    });
```
djhouseknecht, line 48: I didn't actually have to change any of the tests. Just restructured them and removed an old one that no longer applies with my new changes. I guess that is one good thing about writing IT tests, they should still pass when you make some changes :D

Jenkinsfile
```diff
@@ -1,4 +1,4 @@
-@Library('pipeline-library') _
+@Library('pipeline-library@post-release-step-creds-for-github') _
```
djhouseknecht, line 1: I'm testing this feature-branch out ([see diff here](https://bitbucket.org/inindca/pipeline-library/branches/compare/post-release-step-creds-for-github%0Dmaster#diff)). 

Currently, `shouldTagOnRelease` gets a permissions error from github when pushing tags. I think that pipeline-library should fix that issue. See the error at the end of [this build](https://jenkins.ininica.com/job/web-pipeline-deploy/594/consoleFull) for ewt.

jensengar (approved):

PR #90, opened 2021-03-15
[PCM-1557] fix force turn for streaming client

jensengar:


src/webrtc.ts
```diff
@@ -526,10 +521,25 @@ export class WebrtcExtension extends EventEmitter {
...
+
     return iceServers;
   }
 
+  setIceServers (iceServers: any[]) {
+    this.client._stanzaio.jingle.iceServers = iceServers;
+  }
+
+  setIceTransportPolicy (policy: 'relay' | 'all') {
+    this.client._stanzaio.jingle.config.peerConnectionConfig!.iceTransportPolicy = policy;
```
jensengar, line 540: I don't like having to tunnel so deep into jingle, but there doesn't seem to be a better way to do this.

djhouseknecht (approved):
These changes to the streaming-client config aren't breaking are they?

jensengar:
> These changes to the streaming-client config aren't breaking are they?

no they are not.

PR #95, opened 2021-03-23
PCM-1572…

djhouseknecht:
remove retry from notifications POST request to avoid 'stacking' requests during network or API outages

### DIFF
--- a/src/http-client.ts
+++ b/src/http-client.ts
@@ -23,24 +23,22 @@ export class HttpClient {
     504
   ]);
 
-  async requestApiWithRetry (path: string, opts: RequestApiOptions, retryInterval?: number): Promise<any> {
-    const request = retryPromise<any>(
+  requestApiWithRetry<T = any> (path: string, opts: RequestApiOptions, retryInterval?: number): RetryPromise<T> {
+    const retry = retryPromise<T>(
       this.requestApi.bind(this, path, opts),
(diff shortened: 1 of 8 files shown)


src/notifications.ts
```diff
@@ -187,7 +187,7 @@ export class Notifications {
       logger: this.client.logger
     };
     const channelId = this.client.config.channelId;
-    return this.client.http.requestApiWithRetry(`notifications/channels/${channelId}/subscriptions`, requestOptions);
+    return this.client.http.requestApi(`notifications/channels/${channelId}/subscriptions`, requestOptions);
```
djhouseknecht, line 190: This is the main change. Check the ticket for the behavior this was causing.

src/http-client.ts
```diff
@@ -23,24 +23,22 @@ export class HttpClient {
     504
   ]);
 
-  async requestApiWithRetry (path: string, opts: RequestApiOptions, retryInterval?: number): Promise<any> {
-    const request = retryPromise<any>(
+  requestApiWithRetry<T = any> (path: string, opts: RequestApiOptions, retryInterval?: number): RetryPromise<T> {
```
djhouseknecht, line 26: I changed this to return the `RetryPromise` and not just the `Promise` so callers could have more control over if/when to cancel the request. By just returning the promise, the caller would have to wait for the promise to finish (which could be indefinitely).

jensengar (approved):

PR #96, opened 2021-03-24
Remove potential PII from error logging

djhouseknecht:
remove url, responseBody, requestBody, and error message from failed http request as these may contain PII

### DIFF
--- a/src/http-client.ts
+++ b/src/http-client.ts
@@ -91,13 +91,14 @@ export class HttpClient {
         status: error.status,
 
         correlationId: res.headers['inin-correlation-id'],
-        responseBody: res.text,
+        // Potentially could contain PII
+        // responseBody: res.text,
+        // requestBody: res.req._data,
+        // url: res.error.url,
+        message: 'Error making HTTP request', // res.error.message,
 
         method: res.req.method,
-        requestBody: res.req._data,
 
-        url: res.error.url,
-        message: res.error.message,
         name: res.error.name,
         stack: res.error.stack
       };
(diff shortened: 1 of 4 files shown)


src/request-logger.ts
```diff
@@ -4,7 +4,7 @@ export default function attachSuperagentLogger (logger, data, req) {
   let method = req.method;
   logger = logger || console;
 
-  logger.debug(`request: ${method.toUpperCase()} ${req.url}`, { timestamp, data });
+  logger.debug(`request: ${method.toUpperCase()} ${req.url}`, { timestamp, data }, true);
```
djhouseknecht, line 7: I like to be safe with debug logs incase the logging is turned up and debug starts getting sent to the server.

src/http-client.ts
```diff
@@ -91,13 +91,14 @@ export class HttpClient {
...
+        // Potentially could contain PII
+        // responseBody: res.text,
+        // requestBody: res.req._data,
+        // url: res.error.url,
+        message: 'Error making HTTP request', // res.error.message,
 
         method: res.req.method,
-        requestBody: res.req._data,
 
-        url: res.error.url,
```
jensengar, line 99: I would think url and error message should be fine.

djhouseknecht: I had a few thoughts about the url and message: 
1. Query params _could_ contain customer data (like search params)
2. The `error.message` from `superagent` is usually `Failed to POST to ${url}` which is why I removed it (same reason as point 1)
3. This HttpClient class is exported (because the sdk uses), but it isn't publicized. I know we could keep the url since only SC and the SDK will be using it (ie. we know the URL won't contain PII), however, I am just wondering if that could open a pitfall if we ever make a change where the sdk or sc makes a request with customer supplied query params. 

Thoughts?

jensengar: that's fair. and we'd have the correlationId so we could look it up anyway.

jensengar (approved):

PR #97, opened 2021-03-25
[PCM-1561] - Added logging for ICE and connection states.

zservies:


src/webrtc.ts
```diff
@@ -234,11 +234,19 @@ export class WebrtcExtension extends EventEmitter {
     //   this.emit('send', data);
     // });
 
-    this.client._stanzaio.on('jingle:outgoing', (session) => {
+    this.client._stanzaio.on('jingle:outgoing', (session: any) => {
```
djhouseknecht, line 237: NIT: this may be able to be typed as `JingleSession` from `import JingleSession from 'stanza/jingle/Session'`

src/webrtc.ts
```diff
@@ -234,11 +234,19 @@ export class WebrtcExtension extends EventEmitter {
...
       return this.emit(events.OUTGOING_RTCSESSION, session);
     });
 
-    this.client._stanzaio.on('jingle:incoming', (session) => {
+    this.client._stanzaio.on('jingle:incoming', (session: any) => {
```
djhouseknecht, line 241: Same

djhouseknecht (approved):

src/types/media-session.ts
```diff
@@ -31,9 +31,14 @@ export class GenesysCloudMediaSession extends MediaSession {
   }
 
   onIceStateChange () {
-    const state = this.pc.iceConnectionState;
+    const iceState = this.pc.iceConnectionState;
+    const connectionState = this.pc.connectionState;
```
jensengar, line 35: So this is related to the iceConnectionState but isn't exactly the same thing. I would say we should have two separate handlers/logs for them. It looks like stanza doesn't expose the connectionState handler like they do for the ice state so we probably just need to add one in the constructor. Something like this:
```
this.pc.addEventListener('connectionState', this.onConnectionStateChange.bind(this))
```
and we'll remove the connectionState stuff from this fn. Does that make sense?

jensengar (changes requested):

src/types/media-session.ts
zservies: @jensengar Just to verify - you think we should leave the the iceConnectionState how it is and create a handler for connectionState?

jensengar: @zservies not quite. I like what you have here for the iceConnectionState handler stuff, just remove the stuff relating to `pc.connectionState`. Then create a `connectionStateChange` function that basically does the same thing, but for the `pc.connectionState` rather than the `pc.iceConnectionState`.

zservies: @jensengar Yeah thats what I was essentially trying to get at.

djhouseknecht (approved):

PR #98, opened 2021-04-22
[PCM-1547] - Fixed 1:1 video alert toast hanging after answer timeout…

zservies:
… and fixed invites hanging when accepting 1:1 video call with two clients open.


src/webrtc.ts
```diff
@@ -302,6 +306,17 @@ export class WebrtcExtension extends EventEmitter {
     );
   }
 
+  private handleRetract (sessionId: string) {
```
zservies, line 309: As these get longer its probably worth giving them a dedicated stanza handler file.

jensengar (approved):

djhouseknecht (approved):

PR #106, opened 2021-06-10
[PCM-1624] - Added error logging for failed HTTP requests to console - not sending to sumo.

zservies:
- We already aren't sending these HTTP failures to Sumo. Its possible they'll contain PII so instead we'll just throw an error and log them to the console. 
- Removed _sanitized_ error from streaming-client.
- Still exposing `formatRequestError` so that consuming applications can clean up errors before sending to Sumo.

### DIFF
--- a/src/http-client.ts
+++ b/src/http-client.ts
@@ -39,7 +39,6 @@ export class HttpClient {
   }
 
   async requestApi (path: string, opts: RequestApiOptions): Promise<any> {
-    console.error(opts.logger);
     let response = request[opts.method](this._buildUri(opts.host, path, opts.version))
       .use(reqlogger.bind(this, opts.logger, opts.data))
       .type(opts.contentType || 'json');
--- a/src/http-client.ts
+++ b/src/http-client.ts
@@ -58,7 +58,6 @@ export class HttpClient {
         status: err.status,
         correlationId: res.headers['inin-correlation-id'],
         responseBody: res.text,
-        requestBody: res.req._data,
         url: res.error.url,
         message: res.error.message,
         method: res.req.method,
--- a/src/http-client.ts
+++ b/src/http-client.ts
@@ -39,6 +39,7 @@ export class HttpClient {
   }
 
   async requestApi (path: string, opts: RequestApiOptions): Promise<any> {
+    console.error(opts.logger);
     let response = request[opts.method](this._buildUri(opts.host, path, opts.version))
       .use(reqlogger.bind(this, opts.logger, opts.data))
       .type(opts.contentType || 'json');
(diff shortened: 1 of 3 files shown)


src/http-client.ts
```diff
@@ -51,7 +51,20 @@ export class HttpClient {
     try {
       return await response.send(opts.data); // trigger request
     } catch (err) {
-      throw this.formatRequestError(err);
+      const res = err.response;
+
+      // Potentially contains PII information - Do not send to Sumo, just log to console.
+      throw {
```
djhouseknecht, line 57: By just throwing a newly constructed object here, will it still be `instanceof Error`? It may be worth it to just throw the error as is. But it has been a minute this I've messed with superagent errors. Thoughts?

zservies: You're right, I'm going to update this real quick and add the `Error` object to this so that there will be a stack trace.

So it'll be:
```
throw {
someStuff: res.someStuff,
error: new Error()
}
```

zservies: Actually nevermind... we already have a stack trace in the object.

zservies: I like throwing the object because its cleaner to read in the console and this is how it was logged before the Great Log Purge of 2021.

src/http-client.ts
```diff
@@ -51,7 +51,20 @@ export class HttpClient {
...
+        status: err.status,
+        correlationId: res.headers['inin-correlation-id'],
+        responseBody: res.text,
+        requestBody: res.req._data,
```
jensengar, line 61: I'm not sure we can/should log the requestBody as it could contain PII.

zservies: I'll remove it

djhouseknecht (changes requested):
I think there may be some confusion on what this ticket is supposed to do. Right now, we sanitize the error before we throw it. This is super safe knowing no caller (including streaming-client) will ever log PII, but we are trying to change that. I think the caller should be responsible for redacting the error before it logs to sumo. That shouldn't be streaming-client's responsibility. 

How I interpreted this ticket: 
* Streaming-client should _not_ mess with the current error and probably just throw it as is (the superagent error). 
* Streaming-client should expose 1 or 2 methods to format super agent errors better: 
  * `formatRequestError` could format the error the way Zach has in this PR. Just make things nicer to view (do not redact any information). 
  * `formatAndRedactRequestError` (or something similar) could format the error _and_ redact any potential PII fields like request body, url, and response body. 

Consumers could then leverage those utility function before logging anything. 

Currently, the implementation in this PR is formatting the error to a regular object with different fields. Like: 
``` ts
try {
  return await response.send(opts.data); // trigger request
} catch (err) {
  throw { ... };
}
```
But by doing that, we aren't actually throwing an error, but just a normal js object. I personally think it should throw an error and then the caller can use a utility function to format it to a nicer js object if it wants. Right now this would happen: 
``` ts
try {
  await requestApi('some/endpoint', {...}); // if this throws
} catch (error) {
  if (error instanceof Error) {
    console.log('It threw an actual Error object'); // this will never get called
  } else {
    console.log('It threw a POJO as an error'); // this will called
  } 
}
```

jensengar (approved):

djhouseknecht (approved):

PR #110, opened 2021-08-05
PCM-1529 PCM-1701 

djhouseknecht:
fixed soft reconnects. remove google stun servers from defaults. added refreshIceServers retry logic

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -247,14 +247,6 @@ export class Client {
       this[extensionName] = extension.expose;
       this[`_${extensionName}`] = extension;
     });
-
-    /* clear out stanzas default use of google's stun server */
-    this._stanzaio.jingle.config.iceServers = [];
-    /**
-     * NOTE: calling this here should not interfere with the `webrtc.ts` extension
-     *  refreshingIceServers since that is async and this constructor is sync
-     */
-    this._webrtcSessions.setIceServers([]);
   }
 
   cleanupLeakTimer () {
--- a/src/client.ts
+++ b/src/client.ts
@@ -247,6 +247,14 @@ export class Client {
       this[extensionName] = extension.expose;
       this[`_${extensionName}`] = extension;
     });
+
+    /* clear out stanzas default use of google's stun server */
+    this._stanzaio.jingle.config.iceServers = [];
+    /**
+     * NOTE: calling this here should not interfere with the `webrtc.ts` extension
+     *  refreshingIceServers since that is async and this constructor is sync
+     */
+    this._webrtcSessions.setIceServers([]);
   }
 
   cleanupLeakTimer () {
@@ -325,7 +333,8 @@ export class Client {
       const opts: RequestApiOptions = {
         method: 'get',
         host: this.config.apiHost,
-        authToken: this.config.authToken
+        authToken: this.config.authToken,
+        logger: this.logger
       };
       jidPromise = this.http.requestApiWithRetry('users/me', opts).promise
         .then(res => res.body.chat.jabberId);
--- a/src/http-client.ts
+++ b/src/http-client.ts
@@ -34,9 +34,10 @@ export class HttpClient {
     this._httpRetryingRequests.set(retry._id, retry);
 
     /* tslint:disable:no-floating-promises */
-    retry.promise
-      .then(() => this.cancelRetryRequest(retry._id))
-      .catch(() => this.cancelRetryRequest(retry._id));
+    retry.promise.then(
+      () => this.cancelRetryRequest(retry._id),
+      () => this.cancelRetryRequest(retry._id)
+    );
 
     return retry;
   }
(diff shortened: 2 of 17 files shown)


src/reconnector.ts
```diff
@@ -47,7 +47,7 @@ export class Reconnector {
         return;
       }
 
-      if (this.client._stanzaio.transport?.hasStream || this.client.connecting) {
+      if (this.client._stanzaio.transport?.hasStream) {
```
djhouseknecht, line 50: The issue was `this.client.connecting` was set to `true` but stanza tore down the actual websocket. So we were sitting in a state where the reconnector _thought_ we were reconnecting but really we were just waiting for backoff attempts to run out before trying to hard reconnect. 

With this change, it will not drop into this `if` block and will go on to line 58 where stanza tries to create another WS with the existing channel.

src/types/media-session.ts
```diff
@@ -81,7 +81,6 @@ export class GenesysCloudMediaSession extends MediaSession {
       const redactionInfo = loggingOverrides[message];
 
       if (redactionInfo?.skipMessage) {
-        console.debug('not sending log message due to loggingOverride', { message, data });
```
djhouseknecht, line 84: This was causing issues for the pure scale folks.

src/webrtc.ts
```diff
@@ -508,37 +513,67 @@ export class WebrtcExtension extends EventEmitter {
     this.logger.info('sent jingle retract', { sessionId, conversationId: session.propose.conversationId });
   }
 
-  async refreshIceServers (): Promise<any[]> {
+  async refreshIceServers (): Promise<ExtendedRTCIceServer[]> {
+    if (!this.refreshIceServersRetryPromise) {
+      this.refreshIceServersRetryPromise = retryPromise<ExtendedRTCIceServer[]>(
```
djhouseknecht, line 518: Converted this to use a `retryPromise` to retry 3 times if this errors for any reason.

src/client.ts
```diff
@@ -247,6 +247,14 @@ export class Client {
...
       this[`_${extensionName}`] = extension;
     });
+
+    /* clear out stanzas default use of google's stun server */
+    this._stanzaio.jingle.config.iceServers = [];
+    /**
+     * NOTE: calling this here should not interfere with the `webrtc.ts` extension
+     *  refreshingIceServers since that is async and this constructor is sync
+     */
+    this._webrtcSessions.setIceServers([]);
```
jensengar, line 257: we should do this and the line above in `webrtc.ts` as part of its setup.

src/http-client.ts
```diff
@@ -27,18 +27,21 @@ export class HttpClient {
...
-      retryInterval
+      retryInterval,
+      opts.logger
     );
 
     this._httpRetryingRequests.set(retry._id, retry);
 
     /* tslint:disable:no-floating-promises */
-    retry.promise.finally(() => this.cancelRetryRequest(retry._id));
+    retry.promise
```
jensengar, line 37: Is this to add one more retry?

djhouseknecht:
Updated package-lock to version 2 and ensured `async v3.2.1` was installed. the `Jenkins` file uses `npm ci` so we should always get that version. Although, we will need to make sure the sdk, wd, volt, and yak all get the new version too when they bump sc.

src/webrtc.ts
```diff
@@ -508,37 +513,67 @@ export class WebrtcExtension extends EventEmitter {
...
+        this._refreshIceServers.bind(this),
+        (error): boolean => {
+          if (++this.discoRetries > MAX_DISCO_RETRIES) {
+            this.logger.warn('fetching turn servers failed. max retries reached.', {
```
jensengar, line 522: can we change `trun servers` to `ice servers` in these logs please?

src/webrtc.ts
```diff
@@ -508,37 +513,67 @@ export class WebrtcExtension extends EventEmitter {
     this.logger.info('sent jingle retract', { sessionId, conversationId: session.propose.conversationId });
   }
 
-  async refreshIceServers (): Promise<any[]> {
+  async refreshIceServers (): Promise<ExtendedRTCIceServer[]> {
```
jensengar, line 516: I can't add a comment in the place I want, but do you think we should log an explicit error in `on.('connected'...)` if the `refreshIceServers()` fails?

jensengar (changes requested):

src/http-client.ts
jensengar: Ahh ok I see. I would suggest changing this just slightly then because if the `.then()` fails, it's going to try again in the catch which is why I was confused. To do *exactly* what you're saying, I think you'd want to do something like:
`.then(() => this.cancelRetryRequest(retry._id), () => this.cancelRetryRequest(retry._id))` without a `.catch()`

jensengar (approved):

zservies (approved):

PR #114, opened 2021-09-08
[PCM-1715] - Updated streaming-client to end sessions.

zservies:


src/types/media-session.ts
```diff
@@ -89,6 +89,28 @@ export class GenesysCloudMediaSession extends MediaSession {
...
+    if (typeof reason === 'string') {
+      reason = { condition: reason };
+    }
+
+    if (!silent) {
+      this.send('session-terminate', { reason });
+    }
+
+    setTimeout(() => {
+      if (this.pc) {
```
jensengar, line 105: I think this case will always be true. We should check the connection status: https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/connectionState.

We probably care if connectionState is `connecting` or `connected`

jensengar: Can you also add a few tests for this please?

jensengar (changes requested):

src/types/media-session.ts
```diff
@@ -89,6 +89,28 @@ export class GenesysCloudMediaSession extends MediaSession {
...
+
+    if (typeof reason === 'string') {
+      reason = { condition: reason };
+    }
+
+    if (!silent) {
+      this.send('session-terminate', { reason });
+    }
+
+    setTimeout(() => {
```
jensengar, line 104: Can you also add a comment in the code about why we are waiting before we close the peerconnection. I think it'll be helpful when we look at this someday and wonder what the heck we are doing here.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 # [Unreleased](https://github.com/purecloudlabs/genesys-cloud-streaming-client/compare/v13.2.9...HEAD)
+### Added
+* [PCM-1715](https://inindca.atlassian.net/browse/PCM-1715) - Added ability to end sessions from the client.
```
djhouseknecht, line 9: This may be a NIT, but could this note be a little more clear? Technically, we already have the ability to end a session from the client. This PR is more overriding the current stanza implementation of ending sessions, and lets the backend end the PC or it ends it after 2 seconds.

package.json
```diff
@@ -9,7 +9,7 @@
   },
   "main": "dist/npm/index.js",
   "module": "dist/es/module.js",
-  "browser": "dist/streaming-client-browser.js",
+  "browser": "dist/streaming-client.browser.js",
```
djhouseknecht, line 12: It may be worth it (for tracking purposes) to open a PCM ticket to tie this too. That way it will have a history you can tie the commit, version, and dev center issue to. Also, the CHANGELOG should be updated with this change too since it does affect consumption of sc.

djhouseknecht (changes requested):
Just some minor chores for your most recent commit: 
* create a ticket to track the change and dev center forum post
* make sure there is a commit with the newly created ticket (so jira and the CM can find it)
* update the changelog to include the build change (with the new ticket too).

package.json
zservies: Updated.

djhouseknecht (approved):

jensengar (approved):

PR #116, opened 2021-09-16
Prep v13.3.0.

zservies:


jensengar:
usually when I do version bumps, i run `npm version <major|minor|patch> --no-git-tag-version` and it will bump the package.json and package.lock accordingly. I mention that because the lockfile didn't get bumped.

jensengar (approved):

Issue #117, opened 2021-09-23
Browser field is causing build errors with certain building tools


djhouseknecht:
Mainly tested with webpack. It will build just fine, but then when loaded in the browser an error like this occurs: 

`Uncaught TypeError: can't convert undefined to object` 

The issue is the builders are pulling in the built/bundled streaming-client `"browser"` file. 

Resolution is to remove that field because it isn't really needed anyway since streaming-client is already for the browser. 

For internal use: [PCM-1726](https://inindca.atlassian.net/browse/PCM-1726)

djhouseknecht:
This was fixed in v13.3.1.

PR #118, opened 2021-09-27
PCM-1726 

djhouseknecht:
removed package.json 'browser' field and pointed 'module' field to correct file name

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "genesys-cloud-streaming-client",
-  "version": "13.3.0",
+  "version": "13.3.1",
   "description": "client for the Genesys Cloud Streaming APIs (websocket/xmpp interface)",
   "repository": "https:github.com/purecloudlabs/genesys-cloud-streaming-client",
   "license": "MIT",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
   },
   "main": "dist/npm/index.js",
   "module": "dist/es/index.js",
-  "browser": "dist/streaming-client.browser.js",
+  "web": "dist/streaming-client.browser.js",
```
djhouseknecht, line 12: `"web"` does not have any significance for build tools. I renamed it to this just so you can quickly see what the file name and purpose of the file is.

package.json
```diff
@@ -9,7 +9,7 @@
   },
   "main": "dist/npm/index.js",
   "module": "dist/es/index.js",
```
djhouseknecht, line 11: @jensengar can you confirm that this is the file that worked in Web-dir? I think it was (and I am going to test it more just to confirm).

jensengar: i will, but it might not be until tomorrow.

djhouseknecht:
This will fix #117

zservies (approved):

PR #119, opened 2021-09-28
PCM-1727 added parent app name/version to logger appName/Version

djhouseknecht:
@jensengar is this what you had in mind for this one?

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -75,8 +75,8 @@ export interface ClientOptions {
   logger?: any;
   optOutOfWebrtcStatsTelemetry?: boolean;
   allowIPv6?: boolean;
-  parentAppName?: string;
-  parentAppVersion?: string;
+  appName?: string;
+  appVersion?: string;
 }
 
 export class Client {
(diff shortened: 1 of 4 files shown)


maxwellmooney13 (approved):

zservies (approved):

src/client.ts
```diff
@@ -75,6 +75,8 @@ export interface ClientOptions {
   logger?: any;
   optOutOfWebrtcStatsTelemetry?: boolean;
   allowIPv6?: boolean;
+  parentAppName?: string;
```
jensengar, line 78: Do you think it makes sense to just call this `appName` and `appVersion`?

djhouseknecht: Yeah, I think that is better since the consuming app doesn't really care that it is the "parent".

src/webrtc.ts
```diff
@@ -188,9 +188,12 @@ export class WebrtcExtension extends EventEmitter {
       return;
     }
 
+    const appName = `streamingclient${this.client.config.appName ? '-' + this.client.config.appName : ''}`;
+    const appVersion = `${Client.version}${this.client.config.appVersion ? '-' + this.client.config.appVersion : ''}`;
+
     const data = {
```
jensengar, line 194: ok so here's what I had envisioned here:
```
    const data = {
      appName: 'streamingclient',
      appVersion: Client.version,
      parentAppName: this.client.config.appName,
      parentAppVersion: this.client.config.appVersion,
      actions: statsToSend
    };

PR #121, opened 2021-10-12
PCM-1737 pulled in gc-client-logger

djhouseknecht:


src/webrtc.ts
```diff
@@ -207,8 +206,8 @@ export class WebrtcExtension extends EventEmitter {
     const data = {
       appName: 'streamingclient',
       appVersion: Client.version,
-      secondaryAppName: this.client.config.appName,
-      secondaryAppVersion: this.client.config.appVersion,
+      originAppName: this.client.config.appName,
```
jensengar, line 209: does this serve a purpose?

jensengar: yes, sorry, i forgot about the nr stuff. carry on.

PR #124, opened 2021-10-19
PCM-1754 added webrtc logs to consolidate sdk logs

djhouseknecht:


djhouseknecht:
This is blocking [this SDK PR](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/pull/635). 

Also to reiterate the comment on that PR: The streaming-client doesn't log anything around `session-init`. I wasn't sure if we wanted to move that into the streaming-client or not. For now, I left as-is and kept the logging in the SDK.

jensengar:
> This is blocking [this SDK PR](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/pull/635).
> 
> Also to reiterate the comment on that PR: The streaming-client doesn't log anything around `session-init`.

This actually isn't true, the logging happens in a super class of GenesysCloudMediaSession. They don't have a separate sending/sent message, but there is logging for the session init and session accept.

PR #130, opened 2021-12-16
PCM-1773 converted SessionTypes to enum. Add rollup bundle

djhouseknecht:


package.json
```diff
@@ -9,6 +9,7 @@
   },
   "main": "dist/npm/index.js",
   "module": "dist/es/index.js",
+  "rollup:bundle": "dist/es/index.bundle.js",
```
djhouseknecht, line 12: Honestly, this isn't really necessary. But I thought it would be good to have the option to use it. I named it differently than the SDK's so bundlers don't bundle this giant file _and_ the SDK's giant file.

zservies (approved):

jensengar (approved):

PR #146, opened 2022-05-13
[PCM-1910] fix session-reject stanza

jensengar:


djhouseknecht (approved):
I didn't realize that the backend was expecting the id as `sessionId`

jensengar:
> I didn't realize that the backend was expecting the id as `sessionId`

It's not, the jxt definition is just a bit wonky. The object uses `sessionId` but the stanza uses `id`.

PR #147, opened 2022-05-23
Remove topic on last handler unsubscribe

shanebruggeman:
We're encountering a case where our client subscribes to many topics (note that handlers are passed)

```
# call to subscribe ~50 topics at a time, letting streaming client update subscriptions after a reasonable debounce
connection.notifications.subscribe(topic, handler, false, priority).then(...);
```

Later when we're done, the same is done in reverse for `unsubscribe`

```
# call to unsubscribe most of the same topics
connection.notifications.unsubscribe(topic, handler)
```

### Problem
Shortly after unsubscribing the relevant topics, we observed a resubscription to those same topics. Internally the issue is that the set of `subscriptions` and `bulkSubscriptions` remains unchanged upon unsubscribe, which in turn causes a subsequent resubscribe call to think those topics are still required

### Solution
When there are no handlers for a given topic, remove them from the tracked list(s) of subscriptions.

Let me know if I'm off somewhere. I've already run this locally and it's corrected our use case but it's entirely possible that another of your consumers will be affected.

### DIFF
--- a/src/notifications.ts
+++ b/src/notifications.ts
@@ -210,6 +210,10 @@ export class Notifications {
       if (handlerIndex > -1) {
         handlers.splice(handlerIndex, 1);
       }
+      if (!handlers.length) {
+        this.subscriptions[t] = [];
+        delete this.bulkSubscriptions[t];
+      }
     });
   }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 # [Unreleased](https://github.com/purecloudlabs/genesys-cloud-streaming-client/compare/v13.4.1...HEAD)
 
+### Changed
+* [PCM-1862](https://inindca.atlassian.net/browse/PCM-1862) - remove topics from the tracked lists (subscriptions, bulkSubscriptions) after their last handlers have been removed
+
 ### Added
 * [PCM-1837](https://inindca.atlassian.net/browse/PCM-1837) – add `setAccessToken(token)` function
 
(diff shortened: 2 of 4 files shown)


jensengar (changes requested):

djhouseknecht (approved):

src/notifications.ts
```diff
@@ -210,6 +210,10 @@ export class Notifications {
       if (handlerIndex > -1) {
         handlers.splice(handlerIndex, 1);
       }
+      if (!handlers.length) {
+        this.subscriptions[t] = [];
+        delete this.bulkSubscriptions[t];
```
djhouseknecht, line 215: I think I found the root cause. I will try to leave a more detailed explanation, but for clarity, there are two issues with this approach: 

1. By only setting `this.subscriptions[t]` to an empty array, won't actually remove the sub from any future bulk subscribes (more on that in my in detailed explanation). But you will actually want to just delete the whole property
2. We don't want to delete `this.bulkSubscriptions[t]` because it is possible another consumer subscribed to the same topic using the bulk subscribe instead of an individual topic. Unsubscribing from individual topics should not remove bulk subscriptions. 

#### More on bullet 2
The bug is actually in the `resubscribe()` function. SC will eventually treat _all_ individual topic subscriptions as bulk subscriptions. That is why when you unsubscribe from an individual sub, sc never actually removes it because it _also thinks_ that it is a bulk sub so it don't remove it. 

If we can correct the `resubscribe()` function to know the difference between bulk subs and individual subs, we can use part of this to make sure we don't keep subs hanging around. 

I appreciate your work on this, @shanebruggeman. I will open a PR with my fix and I'll have you look it over to see if it fixes your issue.

shanebruggeman: Appreciate the review @djhouseknecht! I'm more than willing to review a better change - this is just what I cobbled together based off following the subscribe / unsubscribe logic. And feel free to close this when it makes sense!

djhouseknecht:
@shanebruggeman I opened #148 if you wanted to take a look at that. I don't know if you can leave comments on the PR or not since you are a "maintainer" of the repo. You can ping me in GC if you can't.

PR #149, opened 2022-05-26
PCM-1908 fixed timing WS out but not actually stopping stanza's connection attempt

djhouseknecht:
## Problem
1. if we timed out `sc.connect()` after the 10 seconds but stanza's WS is still trying to connect, it is possible that i connects _after_ we have rejected with the timeout. So consumers would have already tried to connect another WS. 
1. sometimes our retry on disconnect logic would kick in when stanza ran into an error connecting. This could cause the retry logic in the backgroung to start even though we had already rejected `sc.connect()` 

## Solution
1. when `sc.connect()` times out, we tell stanza to stop any pending connection
2. when `sc.connect()` calls stanza, it listens for `--transport-disconnect` events from stanza and rejects on them (since this is what stanza emits when a WS failed)
3. we do not turn on our retry logic until _after_ successfully connecting.

### DIFF
--- a/src/types/retry-utils.ts
+++ b/src/types/retry-utils.ts
@@ -1,0 +1,143 @@
+import { v4 } from 'uuid';
+import WildEmitter from 'wildemitter';
+
+export interface IRetryConfig<T = any> {
+  /**
+   * Function to attempt until it completes or rejects based on the
+   *  criteria for retrying.
+   */
+  promiseFn: () => Promise<T>;
+  /**
+   * Retry criteria to retry the promise function on failure. Available options:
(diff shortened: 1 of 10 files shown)


src/types/retry-utils.ts
```diff
@@ -0,0 +1,143 @@
+import { v4 } from 'uuid';
+import WildEmitter from 'wildemitter';
+
+export interface IRetryConfig<T = any> {
```
jensengar, line 4: I honestly didn't know we had something like this. I like it.

jensengar (approved):

PR #153, opened 2022-06-03
PCM-1926 adding http body to requests

djhouseknecht:


src/http-client.ts
```diff
@@ -51,9 +51,8 @@ export class HttpClient {
 
     const params: AxiosRequestConfig = {
       method: opts.method,
-
       url,
-
+      data: opts.data,
```
jensengar, line 55: oops

jensengar (approved):

PR #164, opened 2022-08-16
[PCM-1972] add data channel support

jensengar:


src/types/media-session.ts
```diff
@@ -60,6 +61,13 @@ export class GenesysCloudMediaSession extends MediaSession {
     }
     this.pc.addEventListener('connectionstatechange', this.onConnectionStateChange.bind(this));
     this.pc.addEventListener('icecandidateerror', this.onIceCandidateError.bind(this));
+
+    // TODO: REMOVE BECAUSE THIS IS TEMPORARY
+    this.pc.addEventListener('datachannel', this.dataChannelRequested.bind(this));
```
zservies, line 66: This and `dataChannelRequested()` can be removed.

zservies (changes requested):

zservies (approved):

test/unit/media-session.test.ts
```diff
@@ -126,6 +129,28 @@ describe('GenesysCloudMediaSession', () => {
       expect(spy).toHaveBeenCalledWith(JingleAction.SessionInfo, { info: { infoType: JINGLE_INFO_ACTIVE } });
     });
 
+    it('should call setupDataChannel when connected', () => {
+      const parent = new FakeParent();
+
+      const session = new GenesysCloudMediaSession({
+        options: { parent } as any,
+        sessionType: 'softphone',
```
hjon, line 137: I get the impression we're mostly interested in data channels for video. Is that significant enough that it should be restricted by `sessionType`?

hjon: I think I should clarify this. I wondered if there was any danger in a data channel being offered for a `sessionType` that we didn't expect (since we create one if it's been offered)? (In thinking a bit more, I think the danger is low since we're just emitting an event for clients to listen to.)

jensengar: answered via chat. tldr; it doesn't matter in this case.

PR #166, opened 2022-09-13
[PCM-1963] - Added missing error text from switch to Axios from SuperAgent. Added Jon to CODEOWNERS.

zservies:


src/http-client.ts
```diff
@@ -75,6 +76,10 @@ export class HttpClient {
         let status = response.status;
         let correlationId = response.headers?.[correlationIdHeaderName];
         let body = response.data;
+        let error: IAxiosResponseError = {
+          ...res,
+          text: response.request.response
```
zservies, line 81: Please correct me if I'm wrong, but I don't think this will contain PII. This is the missing field that Dave's tests look for.

jensengar, line 81: what's the difference between `response.response` vs `response.request.response`?

jensengar (approved):

src/http-client.ts
zservies: `response.response` doesn't exist on `AxiosError`

hjon (approved):

PR #173, opened 2022-10-20
[PCM-2020] - Reduce amount of logs to console from the data-channel updates.

zservies:


zservies:
I'm open to feedback on this because even this could become noisy in populated conferences. Basically the question is - do we want to log something to console every time someone new speaks/someone stops speaking?

src/types/media-session.ts
```diff
@@ -293,7 +294,10 @@ export class GenesysCloudMediaSession extends MediaSession {
   }
 
   _handleDataChannelMessage (event: MessageEvent) {
-    this._log('debug', 'data channel message received', event);
+    if (event.data !== this.lastDataChannelMessage?.data) {
```
jensengar, line 297: this won't actually catch anything will it? This is reference equality check which will always be falsy.

zservies: > this won't actually catch anything will it? This is reference equality check which will always be falsy.

Actually this is fine because `this.lastDataChannelMessage` is just a reference to `event`.

jensengar (approved):

PR #175, opened 2022-11-10
Pcm 2024

jensengar:
This one is a doozy... Sorry.


jensengar:
Some notes about the changes I made:

- When streamingClient wants to connect an instance of stanza, it will do so by calling `connectionManager.getNewStanzaConnection()`. This call will only resolve if stanza connects successfully. If the connection fails or times out, it will reject.
- streamingClient will keep calling `connectionManager.getNewStanzaConnection()` until it gets a connected instance at which time it will associate that instance as the `activeStanzaInstance` and pass it to all of its extensions so they can setup/react to a new stanza instance.
- once streamingClient has an `activeStanzaInstance`, it will set up a `disconnected` listener against that instance. If the instance becomes disconnected and `streamingClient.disconnect()` was never called, we assume the disconnect was unintentional and we will connect a new instance as if one never existed. (basically start at the first bullet point). This allows us to treat reconnected stanza instances the same as if it was an initial connection.

- Because of the above mentioned bullets, we were able to remove the reconnector extension.

src/connection-manager.ts
```diff
@@ -0,0 +1,132 @@
+// so the idea here is any time stanza gets disconnected, we are effectively going to kill that instance of
+// stanza and never reuse it. We may reuse the config for a give stanza client, but every connection attempt
```
zservies, line 2: Typo

src/connection-manager.ts
```diff
@@ -0,0 +1,132 @@
...
+// will merit its own brand new client. The purpose of this is because the stanza.connect() is not transactional
+// and failed connection attempts trigger `disconnect` events either by timeout or some actual error. This is
+// a problem because it effectively means we can get multiple disconnect events for a single connection attempt
+// which makes trying to reconnect a stanza instance an tricky endeavor because we don't know if a disconnect
+// event applies to a past connection attempt or if it applies to the most recent connection attempt. For this
```
zservies, line 7: `For this`?

src/connection-manager.ts
```diff
@@ -0,0 +1,132 @@
...
+import Logger from 'genesys-cloud-client-logger';
+import { AgentConfig, createClient } from 'stanza';
+import { SASL } from 'stanza/protocol';
+import { v4 } from 'uuid';
+import { IClientConfig } from './types/interfaces';
+import { NamedAgent } from './types/named-agent';
+import SaslError from './types/sasl-error';
+import { parseJwt, timeoutPromise } from './utils';
+
+// reason, we want a vanilla instance of stanza every time we call `connect` so wires don't get crossed.
```
zservies, line 18: This line belongs with the paragraph above.

src/client.ts
```diff
@@ -284,148 +147,250 @@ export class Client {
...
+    }
+
     return timeoutPromise(resolve => {
-      this._stanzaio.once('disconnected', resolve);
+      this.activeStanzaInstance!.once('disconnected', resolve);
       this.autoReconnect = false;
-      this._reconnector.stop(new Error('Cancelling reconnect')); // just in case there is already an active reconnect trying
       this.http.stopAllRetries();
-      this._stanzaio.disconnect();
+      this.activeStanzaInstance!.disconnect();
```
maxwellmooney13, line 233: I didn't know this notation existed but after a quick search, that's pretty cool and good to know!

src/ping.ts
```diff
@@ -38,21 +41,21 @@ export class Ping {
...
+        jid: this.stanzaInstance.jid
       };
       this.client.logger.warn('Missed a ping.', Object.assign({ error: err }, info));
 
       /* if we have reached max number of missed pings, disconnect */
       if (++this.numberOfFailedPings > this.failedPingsBeforeDisconnect) {
-        this.stop();
         this.client.logger.error('Missed too many pings, disconnecting', Object.assign({ numberOfFailedPings: this.numberOfFailedPings }, info));
-        this.client._stanzaio.sendStreamError({ text: 'too many missed pongs', condition: 'connection-timeout' });
+        this.stanzaInstance.sendStreamError({ text: 'too many missed pongs', condition: 'connection-timeout' });
```
maxwellmooney13, line 57: Is "pong" an actual term???

jensengar: lol. frankly i have no idea. makes sense to me though.

src/client.ts
```diff
@@ -284,148 +147,250 @@ export class Client {
...
+
+    this.removeAllListeners(STANZA_DISCONNECTED);
+    this.removeAllListeners(NO_LONGER_SUBSCRIBED);
+
+    // unproxy events
+    if (disconnectedInstance.originalEmitter) {
+      disconnectedInstance.emit = disconnectedInstance.originalEmitter as any;
+    }
+
+    this.activeStanzaInstance = undefined;
```
zservies, line 202: Why set this to `undefined` and not just `false` since we're disconnected?

src/client.ts
```diff
@@ -284,148 +147,250 @@ export class Client {
...
+      } else {
+        const jidRequestOpts: RequestApiOptions = {
+          method: 'get',
+          host: this.config.apiHost,
+          authToken: this.config.authToken,
+          logger: this.logger
+        };
+        jidPromise = this.http.requestApiWithRetry('users/me', jidRequestOpts).promise
+          .then(res => {
+            // TODO: remove
```
zservies, line 364: Remove?

src/ping.ts
hjon: At least for WebSockets, [it is](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.3). 😄

src/ping.ts
```diff
@@ -38,21 +41,21 @@ export class Ping {
 
   private async performPing (): Promise<void> {
     try {
-      await this.client._stanzaio.ping(this.options.jid);
+      await this.stanzaInstance.ping(this.options.jid);
       this.numberOfFailedPings = 0;
```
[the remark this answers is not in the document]
hjon: My bad. 😞 Still getting used to async/await. I got it now (resetting the count when the ping succeeds).

src/client.ts
jensengar: ty.

src/ping.ts
```diff
@@ -1,8 +1,9 @@
 'use strict';
 
 import { Client } from './client';
+import { NamedAgent } from './types/named-agent';
 
-const DEFAULT_PING_INTERVAL = 15 * 1000; // same default as stanza message timeouts
+const DEFAULT_PING_INTERVAL = 14 * 1000;
```
zservies, line 6: Any reason you changed this to 14 from 15? Did it change in stanza?

src/types/offline-error.ts
```diff
@@ -0,0 +1 @@
+export default class OfflineError extends Error {}
```
zservies, line 1: Why create a brand new type here if its just a copy of `Error`?

src/client.ts
```diff
@@ -284,148 +147,246 @@ export class Client {
...
+    this.logger.error('Failed streaming client connection attempt, retrying', additionalErrorDetails, { skipServer: err instanceof OfflineError });
+    return true;
   }
 
-  _tryToConnect () {
-    this.startServerLogging();
-    this.connecting = true;
+  private async makeConnectionAttempt () {
+    if (!navigator.onLine) {
+      throw new OfflineError('Browser if offline, skipping connection attempt');
```
zservies, line 330: Typo -> "is"

zservies: I already asked this in a different comment, but leaving another here. Is there a specific reason you didn't just use `Error` here?

src/types/timeout-error.ts
```diff
@@ -0,0 +1,5 @@
+export class TimeoutError extends Error {
```
zservies, line 1: Same question as the previous error comments. Why not just leave this as `Error` and pass in a message stating that its a `TimeoutError`?

src/ping.ts
jensengar: Yeah, I changed it because *sometimes* the stanza timeout would trigger first and the logs wouldn't indicate a reason why it was disconnected.

src/types/offline-error.ts
jensengar: I did this for type checking. There are cases where I want to handle a timeout differently than just a regular error. 

```
if (err instance of OfflineError) {
  // do something different
}
```

zservies (approved):

maxwellmooney13 (approved):

hjon (approved):

PR #178, opened 2022-12-19
[PCM-2045] fixed various issues with the connection logic change

jensengar:


src/client.ts
```diff
@@ -346,6 +346,14 @@ export class Client extends EventEmitter {
     this.addInateEventHandlers(stanzaInstance);
     this.proxyStanzaEvents(stanzaInstance);
     stanzaInstance.pinger = new Ping(this, stanzaInstance);
+
+    // handle any extension configuration
+    for (const extension of this.extensions) {
+      if (extension.configureNewStanzaInstance) {
+        await extension.configureNewStanzaInstance(stanzaInstance);
```
jensengar, line 353: so this is the solution I came up with for the race condition in spigot. What was happening is spigot was connecting an instance of the sdk then immediately trying to start a screen share session. The problem was some of the config stuff was getting hung up waiting for the ice servers to come back, more specifically the custom stanza definitions. This is ultimately what was causing my spigot problems because it was sending a malformed `mediaUpgrade` presence stanza.

src/webrtc.ts
```diff
@@ -14,7 +14,7 @@ import { isFirefox } from 'browserama';
 import { definitions, Propose } from './stanza-definitions/webrtc-signaling';
 import { GenesysCloudMediaSession, IGenesysCloudMediaSessionParams, SessionEvents } from './types/media-session';
 import { isAcdJid, isScreenRecordingJid, isSoftphoneJid, isVideoJid, calculatePayloadSize, retryPromise, RetryPromise } from './utils';
-import Client from '.';
+import { Client } from './client';
```
jensengar, line 17: this was the bad import causing issues with the sendStats when using the bundled client.

src/webrtc.ts
```diff
@@ -81,7 +81,7 @@ export class WebrtcExtension extends EventEmitter implements StreamingClientExte
   private stanzaInstance?: NamedAgent;
 
   get jid (): string | undefined {
-    return this.client.config.jid;
+    return this.stanzaInstance?.jid;
```
jensengar, line 84: this was causing us issues when sending some of the stanzas because this wasn't always defined.

src/webrtc.ts
```diff
@@ -668,18 +661,18 @@ export class WebrtcExtension extends EventEmitter implements StreamingClientExte
       });
   }
 
-  async _refreshIceServers (): Promise<ExtendedRTCIceServer[]> {
-    if (!this.stanzaInstance) {
+  async _refreshIceServers (stanzaInstance?: NamedAgent): Promise<ExtendedRTCIceServer[]> {
```
jensengar, line 664: we had to change a lot of this ice stuff because we need to configure/set the ice stuff up before we actually set the instance as the activeInstance.

zservies (approved):

PR #186, opened 2023-02-09
[PCM-2042] sdp over xmpp.

jensengar:


src/stanza-definitions/webrtc-signaling.ts
```diff
@@ -74,6 +78,39 @@ const screenStopDefinition: DefinitionOptions = {
...
+  export interface AgentEvents {
+    /* tslint:disable-next-line no-unnecessary-qualifier */
+    'iq:set:genesysWebrtc': Stanzas.ReceivedIQ & { genesysWebrtc: GenesysWebrtcJsonRpcMessage };
+  }
+}
+
+// this allows parsing xml that looks something like this:
+/*
+  <iq xmlns="jabber:client" [other stuff]>
+    <genesys-webrtc xmlns="genesys">{ "id": "whatver", "anyOtherFieldICareAbout": true }</genesys-webrtc>
```
hjon, line 102: I don't really expect this to be a problem, but I wonder if we should use an XMLNS along the lines of `http://genesys.com/...` to avoid any potential conflicts?

I know there are variations of namespaces, but in this case, I'm taking inspiration from [Service Discovery](https://xmpp.org/extensions/xep-0030.html#registrar-protocol), which uses `http://jabber.org/protocol/disco#info`.

jensengar: Miller and I spent some time trying to figure out xmlns and settled on this one. Unless there's a known reason why we *shouldn't*, let's keep it.

maxwellmooney13 (approved):

zservies (approved):

PR #192, opened 2023-03-08
[PCM-2089] fix connection fanout after a connected instance fails

jensengar:


src/client.ts
```diff
@@ -393,15 +414,15 @@ export class Client extends EventEmitter {
...
       if (stanzaInstance) {
         this.logger.error('Error occurred in connection attempt, but after websocket connected. Cleaning up connection so backoff is respected', { stanzaInstanceId: stanzaInstance.id, channelId: stanzaInstance.channelId });
+        this.removeStanzaBoundEventHandlers();
 
-        // unproxy stanza events so we don't try and reconnect
-        (stanzaInstance as any).emit = stanzaInstance.originalEmitter;
-        stanzaInstance.pinger!.stop();
-        stanzaInstance.disconnect();
+        stanzaInstance.pinger?.stop();
+        await (stanzaInstance as unknown as StanzaClient).disconnect();
```
zservies, line 425: I'm assuming the reason there's two types here is because we could have no `stanzaInstance` and `StanzaClient` doesn't support that?

test/scripts/happy-path.js
```diff
@@ -3,10 +3,5 @@ window.client = client;
...
   .then(() => {
-    const personId = client.me.id;
-    client._notifications.subscribe(`v2.users.${personId}.station`);
-
-    // setTimeout(() => {
-    //   client._notifications.subscribe(`v2.users.${personId}.station`);
-    // }, 3000);
-  })
\ No newline at end of file
+    console.log('happy path connected');
```
zservies, line 6: NIT

src/client.ts
jensengar: No the reason here is *crraazzzzyyyy*. It's stanza's fault. bad typing. From stanza's index file:
```
export function createClient(opts: AgentConfig): Agent {
    const client = new Client(opts);
    client.use(Plugins);

    return (client as unknown) as Agent;
}
```

The `Agent` type has a `disconnect()` fn, but it returns void. the `Client` class has a `disconnect()` fn, but it returns a promise. I want to await the promise.

test/scripts/happy-path.js
jensengar: i'm going to leave this.

hjon (approved):

src/client.ts
zservies: Ew.

zservies (approved):

PR #209, opened 2023-10-17
Pcm 2238 bump client logger to fix media helper issue

maxwellmooney13:


src/webrtc.ts
```diff
@@ -437,7 +437,7 @@ export class WebrtcExtension extends EventEmitter implements StreamingClientExte
       });
       this.currentMaxStatSize = desiredMaxStatsSize;
     } catch (err: any) {
-      if (err.status === 413) {
+      if (err.response.status === 413) {
```
maxwellmooney13, line 440: Should I mimic what was done in client logger? Cover both `err.status` and `err.response.status` along with the number and string?

zservies: No, I think we're good here. This is already a number and isn't being changed to a string so I don't think we need to check strings at all.

zservies (approved):

hjon (approved):

PR #216, opened 2023-11-03
[PCM-2262] fix snyk by bumping axios and client logger

jensengar:


src/client.ts
```diff
@@ -310,7 +310,7 @@ export class Client extends EventEmitter {
         error = new Error('Streaming client connection attempted received and undefined error');
       } else if (err.name === 'AxiosError') {
         const axiosError = err as AxiosError;
-        const config = axiosError.config;
+        const config = axiosError.config!;
```
hjon, line 313: I saw warnings about these, too, and didn't know enough context. Do we know there will always be a config?

hjon (approved):

src/client.ts
jensengar: Fair. I'll adjust so we at least dont double blow up.

zservies (approved):

PR #222, opened 2023-12-11
[PCM-2276] fix and expand stats gathering

jensengar:


src/client.ts
```diff
@@ -78,7 +78,8 @@ export class Client extends EventEmitter {
       jidResource: options.jidResource,
       channelId: null as any, // created on connect
       appName: options.appName,
-      appVersion: options.appVersion
+      appVersion: options.appVersion,
+      appId: options.appId
```
jensengar, line 82: turns out we were never adding the appId to the client.config, even though it was already an optional property.

src/stats-formatter.ts
```diff
@@ -1,48 +1,41 @@
...
+}
 
-    Object.assign(details, deepFlatten(e.remoteTracks, `remoteTrack`));
-    delete details.remoteTracks;
-  } else {
-    details = {};
-    if (event.name !== 'failure') {
-      // TODO: log this out when we get genesys-cloud-client-logger in place (allows logging from anywhere)
-    }
+export function formatStatsEvent (event: StatsEvent, extraDetails: FlatObject = {}): InsightAction<{_eventType: string} & FlatObject> {
```
jensengar, line 19: we aren't as restricted about what we can log so I adjusted accordingly, as well as adjusting for the format changes.

src/types/interfaces.ts
```diff
@@ -195,3 +195,64 @@ export type HeadsetControlsRejection = TypedJsonRpcMessage<'headsetControlsRejec
...
+  actionName: 'WebrtcStats';
+  details: InsightActionDetails<T>;
+};
+
+export type InsightActionDetails<K extends { _eventType: string }> = {
+  _eventType: K['_eventType'];
+  /**
+   * This should be an ISO string
+   */
+  _eventTimestamp: string;
```
jensengar, line 223: Ok, so this is the field we will look at nr for the actual time of the event. We can't actually set the one NR relies on by default. NR will always set it to the date the custom event was received (yes, it's stupid).

src/webrtc.ts
```diff
@@ -226,6 +227,7 @@ export class WebrtcExtension extends EventEmitter implements StreamingClientExte
     const session = new GenesysCloudMediaSession(this, mediaSessionParams);
 
     await session.setRemoteDescription(params.sdp);
+    this.proxyStatsForSession(session);
```
jensengar, line 230: this was a whoopsie. Turns out we weren't sending stats for sdpViaXMPP.

zservies (approved):

maxwellmooney13 (approved):

PR #230, opened 2024-02-02
[PCM-2304] fix backoff

jensengar:


package.json
```diff
@@ -56,12 +56,12 @@
     "browserama": "^3.2.0",
     "core-js": "^3.6.5",
     "debounce-promise": "^3.1.2",
-    "exponential-backoff": "^3.1.0",
+    "exponential-backoff": "^3.1.1",
     "genesys-cloud-client-logger": "^4.2.7",
     "limiter": "^1.1.0",
     "lodash.throttle": "^4.1.1",
     "lru-cache": "^6.0.0",
-    "stanza": "^12.17.2",
```
hjon, line 64: Unless I'm missing something, I think we can bump to stanza 12.20.0. That includes the `queueMicrotask` fix.

index.html
```diff
@@ -26,7 +26,7 @@
       console.log('script loaded    ', src);
     };
 
-    // loadScript('test/scripts/happy-path.js');
+    loadScript('test/scripts/happy-path.js');
```
maxwellmooney13, line 29: Are the changes in this file for test purposes?

webpack.config.js
```diff
@@ -65,7 +65,9 @@ module.exports = (env = {}) => {
       libraryExport: 'default',
       path: Path.resolve(versionDir)
     },
-
+    optimization: {
+      minimize: false
```
hjon, line 69: I feel a little ignorant, but is this an intentional change or was it just for development?

src/client.ts
```diff
@@ -416,6 +514,15 @@ export class Client extends EventEmitter {
...
     try {
+      // TODO: REMOVE
+      // if ((window as any).delayAmount) {
+      //   await delay((window as any).delayAmount);
+      // }
+
+      // TODO: REMOVE
+      // if ((window as any).forceFail) {
+      //   throw new Error();
+      // }
```
hjon, line 525: Given the `REMOVE` comments, I'm guessing these can be removed now?

src/types/interfaces.ts
```diff
@@ -261,3 +261,14 @@ export type MediaStat = InsightAction<{
 
 // This will be a union of all the stats we want to proxy
 export type NRProxyStat = FirstAlertingConversationStat | MediaStat;
+
+export type SCConnectionData = {
```
hjon, line 265: I think this is pretty minor, but I noticed that the order of these fields in `increaseBackoff` seemed to make more sense to me. So I propose changing the order here to match, as follows:

```
currentDelayMs: number;
delayMsAfterNextReduction?: number;
// At this time, we will reduce the current delay.
// This is a long date
nextDelayReductionTime?: number;
// At this time, we will disregard any saved delays. It should be updated every failure.
// This is a long date
timeOfTotalReset?: number;
```

I think it is worth making them consistent, even if we decide a different order is better.

webpack.config.js
jensengar: This was just for dev, good catch.

src/client.ts
```diff
@@ -279,6 +283,79 @@ export class Client extends EventEmitter {
...
+    const msUntilNextReduction = newAmountMs * BACKOFF_DECREASE_DELAY_MULTIPLIER;
+    const newConnectionData: SCConnectionData = {
+      currentDelayMs: newAmountMs,
+      delayMsAfterNextReduction: newAmountMs / 2,
+      nextDelayReductionTime: new Date().getTime() + (msUntilNextReduction),
+      timeOfTotalReset: data.timeOfTotalReset
+    };
+
+    // if we are past the total reset time, do that instead
+    if (data.timeOfTotalReset && data.timeOfTotalReset < new Date().getTime() || newAmountMs < 2000) {
```
jensengar, line 332: put this initial delay in a var

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 # [Unreleased](https://github.com/purecloudlabs/genesys-cloud-streaming-client/compare/v17.0.1...HEAD)
+### Fixed
+* [PCM-2304](https://inindca.atlassian.net/browse/PCM-2304) Made connection backoff semi-persistent through the sessionStore. Updated to forked version of stanza to circumvent browser intensive throttling on connection attempts.
```
hjon, line 9: Nit: We're no longer forking stanza. 😄

hjon (approved):

OjSahin (approved):
Looks good from a QA perspective.

PR #238, opened 2024-03-27
[PCM-2344] - Added ability to pass in custom headers for telemetry purposes.

zservies:


zservies:
I didn't update the documentation with the new `customHeaders` field because this is really only for internal Genesys apps.

jensengar (approved):

PR #250, opened 2024-06-07
[STREAM-96] Forcefully disconnect the WebSocket when missed pings are detected to help reconnect faster

hjon:


src/ping.ts
```diff
@@ -60,6 +60,10 @@ export class Ping {
         this.client.logger.error('Missed too many pings, disconnecting', Object.assign({ numberOfFailedPings: this.numberOfFailedPings }, info));
         this.stanzaInstance.sendStreamError({ text: 'too many missed pongs', condition: 'connection-timeout' });
         this.stop();
+        if (this.stanzaInstance.transport) {
+          this.stanzaInstance.transport.hasStream = false;
```
jensengar, line 64: why do we have to set this?

hjon: Based on the [`disconnect` method for `WSConnection`](https://github.com/legastero/stanza/blob/829bdbf495337f006fd3e9505cf00a8e72b4c98e/src/transports/websocket.ts#L123-L128), it looks like this should be set to `false` if we disconnect uncleanly, so I'll remove this.

jensengar (approved):

zservies (approved):

PR #255, opened 2024-06-21
[STREAM-108] B - Depend on `ws` directly

hjon:
So we can ensure an older version (that's vulnerable) won't be installed.


hjon:
I'm going to just add this directly to the release branch.

PR #260, opened 2024-08-08
[STREAM-69] Don't send stats if browser is offline

hjon:


src/webrtc.ts
```diff
@@ -453,11 +453,17 @@ export class WebrtcExtension extends EventEmitter implements StreamingClientExte
     return logDetails;
   }
 
-  flushStats () {
+  sendStatsImmediately () {
+    // `throttledSendStats` needs to have a scheduled exeuction for `flush` to invoke the throttled function
+    this.throttledSendStats();
     this.throttledSendStats.flush();
```
hjon, line 459: When `throttledSendStats` is called, it starts a timer for the next execution. When `flush` is called, it checks if a timer already exists. If so, it triggers an execution immediately. But if there isn't a timer, it just returns the last results of the throttled function.

See https://github.com/lodash/lodash/blob/4.1.1-npm-packages/lodash.throttle/index.js#L221-L223

jensengar (approved):

zservies (approved):

PR #265, opened 2024-09-17
[STREAM-178] clean up persistent connection session with dead peerCon…

jensengar:
…nections after system wake (genesys sessions only)

### DIFF
--- a/src/types/genesys-cloud-media-session.ts
+++ b/src/types/genesys-cloud-media-session.ts
@@ -21,6 +21,7 @@ export class GenesysCloudMediaSession {
   private iceCandidatesReceivedFromPeer = 0;
   private interruptionStart?: Date;
   private logger: Logger;
+  private stateSyncTimeout?: any;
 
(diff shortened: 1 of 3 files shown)


src/types/genesys-cloud-media-session.ts
```diff
@@ -21,6 +21,7 @@ export class GenesysCloudMediaSession {
   private iceCandidatesReceivedFromPeer = 0;
   private interruptionStart?: Date;
   private logger: Logger;
+  private stateSyncTimeout?: any;
```
maxwellmooney13, line 24: Might be a NIT due to the severity of this but we were trying to move away from `any` right?

jensengar: I messed with it but I was getting compiler errors when running the tests. `NodeJS.Timers` can be annoying to deal with so I just made it an `any`.

maxwellmooney13 (approved):

PR #267, opened 2024-09-20
[STREAM-162] Streaming-client should trigger re-auth on SASL errors

hjon:


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 # [Unreleased](https://github.com/purecloudlabs/genesys-cloud-streaming-client/compare/v17.2.5...HEAD)
+### Breaking Changes
```
jensengar, line 8: I don't think this is technically a breaking change. The contract hasn't really changed. We are still returning an error, we just decorated the error, right?

jensengar (approved):

CHANGELOG.md
hjon: We used to be throwing whatever error we caught here. And some of those errors extend `Error`, like `AxiosError`, which also has the following fields:

```
config?: InternalAxiosRequestConfig<D>;
code?: string;
request?: any;
response?: AxiosResponse<T, D>;
isAxiosError: boolean;
status?: number;
toJSON: () => object;
cause?: Error;
```

But now I've encapsulated those inside a `StreamingClientError`. As a result, if anyone else had been inspecting these errors to look for an auth error with something like `error.response?.status`, they will now have to use something like `error.details.response?.status`.

hjon: Or just look at the type of `StreamingClientError`. 🙂

maxwellmooney13 (approved):

PR #270, opened 2024-10-03
[STREAM-207] handle ice candidates received before the offer (sdpOverXmpp only)

jensengar:


zservies (approved):

hjon:
I realize this isn't Jingle, but even with just SDP, my initial reaction is that having candidates without an offer first isn't normal. 🙂

Maybe I overlooked something or just don't understand the context; could you explain that further?

hjon (approved):

PR #274, opened 2024-11-14
[STREAM-262] GenesysCloudMediaSession does not transition to `ended` when a call ends

hjon:


src/types/genesys-cloud-media-session.ts
```diff
@@ -305,6 +303,8 @@ export class GenesysCloudMediaSession {
   }
 
   onSessionTerminate (reason?: JingleReasonCondition) {
+    this.state = 'ended';
+
     if (this.peerConnection) {
```
hjon, line 308: Not directly related to this ticket, but from the interface (and a bit of code spelunking), it looks like we should always have a peerConnection. Is there a situation I overlooked where we might not have one?

jensengar (approved):

zservies (approved):

hjon:
Merged manually

PR #276, opened 2025-01-08
[STREAM-313] - De-duplicate reinvte offers. Update CODEOWNERS.

zservies:


test/unit/webrtc.test.ts
```diff
@@ -1951,7 +1951,52 @@ describe('handleGenesysOffer', () => {
...
+        params: {
+          sessionId: 'session24',
+          conversationId: 'cid',
+          sdp: 'a=ice-ufrag:asdfasdf\r\na=ice-pwd:fdsafdsa\r\n',
+          reinvite: true
+        }
+      }
+    };
+    webrtc.on('incomingRtcSession', (session: GenesysCloudMediaSession) => {
+      expect(session.setRemoteDescription).toHaveBeenCalled();
```
hjon, line 1972: Checking my understanding of the execution flow: do these have to be verified here or could they also be verified after `handleGenesysOffer` is called?

hjon (approved):

maxwellmooney13 (approved):

src/webrtc.ts
```diff
@@ -89,6 +89,7 @@ export class WebrtcExtension extends EventEmitter implements StreamingClientExte
   private iceServers: RTCIceServer[] = [];
   private stanzaInstance?: NamedAgent;
   private webrtcSessions: GenesysCloudMediaSession[] = [];
+  private previousReinvite: string | undefined = '';
```
hjon, line 92: I had a couple of thoughts today:

1. Do we need to support multiple sessions being re-invited? I imagine the most likely scenario for this is someone with an active call and a call on hold. I don't think we support re-invite for softphone right now, though, so maybe it doesn't matter at this point.
2. Do you think it'd be worth clearing it when we establish the session? I don't think it's a big deal since it's just a single ID that will be overwritten, but it seems like the "right" thing to do. 🙂

CODEOWNERS
```diff
@@ -1 +1 @@
-* @jensengar @zservies @maxwellmooney13 @hjon
```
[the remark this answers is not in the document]
zservies: [*]

PR #280, opened 2025-01-22
[STREAM-326] streaming client is emitting a pendingSession event for jingle and json-rpc

hjon:
This reverts the code from https://github.com/purecloudlabs/genesys-cloud-streaming-client/pull/277 that changed the behavior around deleting pending sessions, which caused the bug where both the Jingle handler and the SDP over XMPP handler emitted sessions.

The other parts of #277 are still valid.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * [STREAM-207](https://inindca.atlassian.net/browse/STREAM-207) - [STREAM-207] handle ice candidates received before the offer (sdpOverXmpp only)
+* [STREAM-326](https://inindca.atlassian.net/browse/STREAM-326) - Only delete pending sessions when handling Jingle session-initiate to avoid emitting two sessions (partial reversion of d1251b22c1b82ddd0b514a5384a89510082dda26).
 
 # [v18.0.0](https://github.com/purecloudlabs/genesys-cloud-streaming-client/compare/v17.2.7...v18.0.0)
 ### Breaking Changes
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * [STREAM-207](https://inindca.atlassian.net/browse/STREAM-207) - [STREAM-207] handle ice candidates received before the offer (sdpOverXmpp only)
+* [STREAM-326](https://inindca.atlassian.net/browse/STREAM-326) - Only delete pending sessions when handling Jingle session-initiate to avoid emitting two sessions (partial reversion of d1251b22c1b82ddd0b514a5384a89510082dda26).
```
hjon, line 19: @zservies @maxwellmooney13 Thoughts on this? I was a little inclined to change the note on PCM-2081 above since this hasn't released yet (it seems less confusing in the log to avoid it altogether), but on the other hand, that would lose the ticket reference. This seemed like a reasonable compromise, but I'm open to ideas.

test/unit/session-initiation-timing.test.ts
```diff
@@ -0,0 +1,130 @@
...
+import { WebrtcExtension } from "../../src/webrtc";
+import { SessionOpts } from "stanza/jingle/Session";
+
+jest.mock('../../src/types/stanza-media-session');
+StanzaMediaSession.prototype.on = jest.fn();
+
+jest.mock('../../src/types/genesys-cloud-media-session', () => {
+  return {
+    GenesysCloudMediaSession: jest.fn().mockImplementation((webrtcExtension, params) => ({
+      meetingId: params.meetingId,
```
hjon, line 20: I tried to put these tests in webrtc.test.ts, but this mock caused issues with other tests, though I'm not quite sure why.

There might be another way to do this, but I wanted a way to verify that we still had a pending session for SDP over XMPP, regardless of the order we handled things in. In the context I've been working in, one way to do this is to make sure the meetingId is grabbed from the pending session and set on the session. But that happens in the constructor, which is why I had to mock this differently.

Let me know if you see another way to accomplish this.

maxwellmooney13 (approved):

zservies (approved):

PR #282, opened 2025-02-13
[STREAM-357] - Maintain JID across hard reconnects/stanza instances from the same client.

zservies:


src/client.ts
```diff
@@ -643,19 +646,21 @@ export class Client extends EventEmitter {
...
-          host: this.config.apiHost,
-          authToken: this.config.authToken,
-          logger: this.logger,
-          customHeaders: this.config.customHeaders
-        };
-        jidPromise = this.http.requestApi('users/me', jidRequestOpts)
-          .then(res => res.data.chat.jabberId);
+      // Use stored JID if we have one, otherwise use the provided JID or grab one.
+      if (!this.initialJid) {
+        if (this.config.jid) {
```
zservies, line 651: I'm not sure if this is actually needed, it may be redundant because we end up assigning `this.config.jid = this.initialJid` later too.

zservies: I think the behavior is mostly the same as it was before.

The idea was - if we don't have a saved JID, then we would try to use the one in the config, but if we don't have one there then we'd fetch one.

And if we fetched one we needed to assign it to the config/saved jid. I'm not sure there's _harm_ in it as it is, but its definitely a bit confusing.

src/client.ts
```diff
@@ -643,19 +646,21 @@ export class Client extends EventEmitter {
     }
 
     if (this.hardReconnectRequired) {
-      let jidPromise: Promise<any>;
-      if (this.config.jid) {
```
zservies, line 647: In the previous code, we would automatically resolve our promise if a JID was passed in via config. What's weird though is we also grabbed the value from the promise and assigned it to `config.jid` even though that value is the same if we end up resolving without needing to fetch via api.

src/client.ts
maxwellmooney13: Yeah I'm following the intent here and I think you're right. I don't think there are any issues as is but it is strange.  I'll still approve it because I think we should still be fine but out of curiosity, would there be any benefit to `this.config.jid = this.initialJid` being moved into the else block here?  Spitballing more than anything and definitely not saying it's a necessary change, just curious more than anything.

src/client.ts
maxwellmooney13: Huh...

maxwellmooney13 (approved):

hjon (approved):

PR #285, opened 2025-03-13
[STREAM-391] Migrate to webappPipelineV2

hjon:


pipeline.d/deploy.yml
```diff
@@ -0,0 +1,54 @@
...
+    - type: habitat
+      name: test
+      schedule: always
+    - type: environment
+      name: test
+      filters:
+        family: webui
+      schedule: always
+      sequence: parallel
+  changeManagement:
```
hjon, line 52: I just copied from what we had elsewhere, but I think we talked about submitting manually; do we still want these values?

package.json
```diff
@@ -25,7 +25,7 @@
     "compile": "tsc -p .",
     "compile:module": "tsc -p . --outDir ./dist/es --target es2015 --module es2015",
     "compile:rollup": "rollup -c rollup.config.js",
-    "compile:webpack": "webpack --mode production",
+    "compile:webpack": "NODE_OPTIONS=--openssl-legacy-provider webpack --mode production",
```
hjon, line 28: I added details in the ticket, but the short version is that webpack v4 uses MD4 for hashing and versions of Node newer than 16 include OpenSSL, which doesn't allow that by default.

Jenkinsfile
```diff
@@ -202,6 +149,15 @@ VERSION      : ${env.VERSION}
             }
         } // end publish to npm
 
+        // Run PureScale when we have a new `next` version
```
hjon, line 152: @maxwellmooney13 Zach and I talked about this, but since you missed it: Wes gave the suggestion for us to trigger the PureScale tests whenever we have a `next` version so we can get that ball rolling a little faster than we do right now.

pipeline.d/service.yml
```diff
@@ -0,0 +1,17 @@
+---
+service:
+  family:
+    - core
+    - webui
+  name: developercenter-cdn--streaming-client-webui
```
hjon, line 6: This seems to match the naming we have right now (and I found out that `/` isn't allowed in this field), but I feel a little wary. 🙂

Jenkinsfile
```diff
@@ -202,6 +149,15 @@ VERSION      : ${env.VERSION}
             }
         } // end publish to npm
 
+        // Initiate PureScale tests when we have a new `next` version
+        if (isRelease()) {
```
zservies, line 153: 🔥

zservies (approved):

maxwellmooney13 (approved):

hjon:
Our approach to pipeline stuff is changing; closing this in light of that.

PR #290, opened 2025-04-17
[STREAM-462] - Update stanza timeout timer to align with server.

zservies:


src/server-monitor.ts
```diff
@@ -3,7 +3,7 @@
 import { Client } from './client';
 import { NamedAgent } from './types/named-agent';
 
-const DEFAULT_STANZA_TIMEOUT = 70 * 1000;
+const DEFAULT_STANZA_TIMEOUT = 35 * 1000;
```
zservies, line 6: Hawk's timer was halved from 30 seconds to 15 so I've halved our timeout as well to be conservative and align.

maxwellmooney13 (approved):

hjon (approved):

PR #291, opened 2025-04-22
[STREAM-85] Handle `v2.system.socket_closing` message

hjon:


src/client.ts
```diff
@@ -280,6 +289,15 @@ export class Client extends EventEmitter {
     this.hardReconnectRequired = true;
   }
 
+  private handleSocketClosing (stanzaInstance: NamedAgent) {
+    this.logger.warn('socket_closing event received', { stanzaInstanceId: stanzaInstance.id, channelId: stanzaInstance.channelId });
+
+    // The docs says we have up to one minute to disconnect and connect a new WebSocket.
+    // So unlike a duplicate channel, we should be proactive and not wait to be disconnected.
+    void this.disconnect();
+    void this.connect({ keepTryingOnFailure: true });
```
hjon, line 298: I don't think we need to await either of these:
- When we `connect`, we'll create a new Stanza instance, so we don't need to wait for the old one to disconnect
- If something happens during the `connect`, we won't be able do much here and will have to rely on the other reconnection handling and events

If you see a reason to do differently, let me know.

Otherwise, is there a more idiomatic way to do this? (I added the `void` because of listing.)

hjon:
Closing for now because this needs to be handled differently.

PR #292, opened 2025-04-29
[STREAM-461] - Remove conferenceId support from JWT auth.

zservies:
Since we updated the backend, we now have a jid present in the JWT for `collaborateVideo` sessions and no longer need the `conferenceId`. This was never released so its not a breaking change.


package.json
```diff
@@ -20,7 +20,7 @@
   "types": "dist/es/index.d.ts",
   "scripts": {
     "prebuild": "rimraf dist",
-    "build": "ts-node scripts/build",
+    "build": "cross-env NODE_OPTIONS='--openssl-legacy-provider' ts-node scripts/build",
```
zservies, line 23: I had trouble building even on node v20 without using `--openssl-legacy-provider`.

zservies: Per this ticket - I suppose I should remove this.
 https://inindca.atlassian.net/browse/STREAM-395?atlOrigin=eyJpIjoiYTgzMjAwYzNhYWM0NGRkNGE0OTI5OTc5NDVkODg0YjAiLCJwIjoiaiJ9

hjon: We won't be able to remove until we move to Webpack 5, but there is an alternative way to do this in the short-term that I used when I ran into this:

`"compile:webpack": "NODE_OPTIONS=--openssl-legacy-provider webpack --mode production"`

My initial reaction is that I prefer this since it keeps it close to where it actually matters (using webpack) and doesn't add a dependency, but this is short-term, anyway, so I don't have strong feelings on this.

PR #293, opened 2025-04-30
[STREAM-518] - Update stanza sent when sending the extension discovery request for agent-video to include full JID from JWT.

zservies:
…y request for agent-video to include the full JID from the JWT.


zservies:
Don't think this is actually needed anymore. Closing.

PR #294, opened 2025-05-05
[STREAM-523] - Properly emit `connectionState` event

maxwellmooney13:


src/types/genesys-cloud-media-session.ts
```diff
@@ -229,6 +229,8 @@ export class GenesysCloudMediaSession {
         this.onSessionTerminate();
       }
     }
+
+    this.emit('connectionState', connectionState);
```
zservies, line 233: Do we not emit this at all anywhere else? I would've figured we did.

zservies (approved):

src/types/genesys-cloud-media-session.ts
maxwellmooney13: There's a chance we _might_ but it didn't look like we were specifically from the session object which is what is being listened to in a few places.  SDK has a check for `session.on(...)` as does Troubleshooter. From what I can tell from testing, we are not emitting the event specifically but we do update the `connectionState` value on the session object if that makes sense. All that is fine but we then had nothing for the troubleshooter to react to.

maxwellmooney13: There is a spot in the Streaming Client `client.ts` file but that seems to proxying Stanza events? I'm not sure if what we're looking for would fall in there OR if we would be even using it while on SDP over XMPP.  With SDP over XMPP off however, that emit comes from Stanza itself (which then in turn I suppose passes through the above Stanza proxy).  But yeah elsewhere throughout Streaming Client had no mention of this specific emit

hjon (approved):

PR #297, opened 2025-05-07
[STREAM-85] Streaming client should reconnect when receiving a message on `v2.system.socket_closing`

hjon:
I've added some notes to the ticket and comments, but Hawk maps the `v2.system.socket_closing` event to [XEP-0051](https://xmpp.org/extensions/xep-0051.html).

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -33,7 +33,6 @@ let extensions = {
 };
 
 const STANZA_DISCONNECTED = 'stanzaDisconnected';
-const SOCKET_CLOSING = 'notify:socket_closing';
 const NO_LONGER_SUBSCRIBED = 'notify:no_longer_subscribed';
 const DUPLICATE_ID = 'notify:duplicate_id';
 const MAX_CHANNEL_REUSES = 10;
@@ -64,7 +63,6 @@ export class Client extends EventEmitter {
   private boundStanzaDisconnect?: () => Promise<any>;
   private boundStanzaNoLongerSubscribed?: () => void;
   private boundStanzaDuplicateId?: () => void;
-  private boundStanzaSocketClosing?: () => void;
 
   http: HttpClient;
   notifications!: NotificationsAPI;
@@ -183,12 +181,10 @@ export class Client extends EventEmitter {
     this.boundStanzaDisconnect = this.handleStanzaDisconnectedEvent.bind(this, stanza);
     this.boundStanzaNoLongerSubscribed = this.handleNoLongerSubscribed.bind(this, stanza);
     this.boundStanzaDuplicateId = this.handleDuplicateId.bind(this, stanza);
-    this.boundStanzaSocketClosing = this.handleSocketClosing.bind(this, stanza);
 
     this.on(STANZA_DISCONNECTED, this.boundStanzaDisconnect);
     this.on(NO_LONGER_SUBSCRIBED, this.boundStanzaNoLongerSubscribed);
     this.on(DUPLICATE_ID, this.boundStanzaDuplicateId);
-    this.on(SOCKET_CLOSING, this.boundStanzaSocketClosing);
 
     this.extensions.forEach(extension => {
       if (typeof extension.handleIq === 'function') {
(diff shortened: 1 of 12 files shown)


src/stanza-definitions/webrtc-signaling.ts
hjon, line 1: As I was working on this, I found out that `AgentEvents` are actually just in the `stanza` module, not `stanza/protocol`, so I changed this (and in webrtc.ts) so we could get rid of the `any` and disabling of tslint.

src/webrtc.ts
hjon, line 1: As I was working on this, I found out that `AgentEvents` are actually just in the `stanza` module, not `stanza/protocol`, so I changed this (and in webrtc-signaling.ts) so we could get rid of the `any` and disabling of tslint.

src/client.ts
```diff
@@ -612,6 +613,8 @@ export class Client extends EventEmitter {
   }
 
   private async setupConnectionMonitoring (stanzaInstance: NamedAgent) {
+    stanzaInstance.connectionTransfer = new ConnectionTransfer(this, stanzaInstance);
```
hjon, line 616: This isn't the same as client-side or server-side pings, but it felt related enough that I thought it made sense to create this in the same function.

src/stanza-definitions/webrtc-signaling.ts
zservies: You're the man.

src/stanza-definitions/xep0051.ts
```diff
@@ -0,0 +1,45 @@
+import { Stanzas } from 'stanza';
+import { childText, DefinitionOptions } from 'stanza/jxt';
+
+// XEP-0051: Connection Transfer
+// Source: https://xmpp.org/extensions/xep-0051.html
```
zservies, line 5: Should we instead point this to RFC 6120 since it now supersedes XEP-0051?

zservies (approved):

src/stanza-definitions/xep0051.ts
hjon: As followup to this, the relevant section is 4.9.3.19: https://www.rfc-editor.org/rfc/rfc6120#section-4.9.3.19. The details for that are different (e.g. it’s a stream error of see-other-host), and Hawk still explicitly uses XEP-0051 at this time.

zservies (approved):

doc/documentation.md
```diff
@@ -42,29 +42,9 @@ one example of a disconnect that will not trigger a reconnect.
...
-draining for a deployment, or load balancing. The client has the ability to delay that reconnect up
-to 10 minutes to allow the application to finish any active task which would be disrupted by a reconnect.
-To subscribe to this event, use the `requestReconnect` event name, which emits with a callback. The
-callback accepts an object with either `pending` or `done` property indicating that the reconnect should
-be delayed or can proceed. If the callback is not called within one second, the disconnect will proceed.
-If the callback is called with `{ pending: true }` then the disconnect will proceed after the callback
-is called with `{ done: true }` or 10 minutes, whichever is sooner. Example:
-
-```js
-client.on('requestReconnect', function (callback) {
```
hjon, line 55: This event was removed in v15.0.0.

doc/documentation.md
```diff
@@ -90,9 +70,6 @@ client.on('requestReconnect', function (callback) {
 `client.connect() : Promise<void>` - Initialize the WebSocket connection for streaming
 connectivity with Genesys Cloud. `connect` must be called before any events will trigger.
 
-`client.reconnect() : Promise<void>` - Disconnect (if connected) and reconnect to
```
hjon, line 93: This was also removed in v15.0.0.

doc/documentation.md
```diff
@@ -42,29 +42,9 @@ one example of a disconnect that will not trigger a reconnect.
...
-    // e.g., an event when all calls have disconnected
-    this.on('canReconnectSafely', function () {
-      callback({ done: true });
-    });
-  }
-});
-```
+- The websocket host service may periodically request the client to reconnect to facilitate
+connection draining for a deployment, or load balancing. If this occurs, the client will
+automatically reconnect.
```
zservies, line 47: Might be worth saying "the client will automatically _attempt_ to reconnect."

PR #298, opened 2025-05-15
[WIP] [STREAM-534] - Added in map of session IDs and whether or not they ae using SDP over XMPP upon propose

maxwellmooney13:


maxwellmooney13:
Opening this as a WIP for now to open up discussions

src/webrtc.ts
```diff
@@ -385,7 +386,7 @@ export class WebrtcExtension extends EventEmitter implements StreamingClientExte
   }
 
   prepareSession (options: SessionOpts): StanzaMediaSession | undefined {
-    if (this.sdpOverXmpp) {
+    if (options.sid && this.sessionsMap[options.sid]) {
```
maxwellmooney13, line 389: Originally we had a check here for `this.sdpOverXmpp` but the issue that was found on the corresponding ticket is that value has the potential to change when it should not.  Softphone call comes in and is using SDP over XMPP, `this.sdpOverXmpp` is set to true. Screen recording comes in and is _not_ using SDP over XMPP, `this.sdpOverXmpp` is set to false and coming to this point, results in this function not getting skipped.

src/webrtc.ts
```diff
@@ -613,7 +614,11 @@ export class WebrtcExtension extends EventEmitter implements StreamingClientExte
    * Stanza Handlers
    */
   private async handlePropose (msg: ProposeStanza) {
-    if (msg.from === this.jid) {
+    // if (msg.from === this.jid) {
+    //   return;
+    // }
```
maxwellmooney13, line 619: We had this condition originally... It's currently commented out because I'm actually not sure if it is needed or not with the change we are planning on making.  I spoke with Jon and we sorta got to the point of "well it probably doesn't _hurt_" so definitely open to suggestions and input here.

src/webrtc.ts
```diff
@@ -95,7 +95,8 @@ export class WebrtcExtension extends EventEmitter implements StreamingClientExte
     max: 5,
     ttl: 1000 * 60 * 3
   });
-  private sdpOverXmpp = false;
+  // private sdpOverXmpp = false;
+  private sessionsMap = {};
```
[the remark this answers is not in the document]
maxwellmooney13: Talking with Jon, it we may need to make the key something like `[sessionType] - [sessionId]` as the session IDs may be the same for both softphone and screen recording.

src/webrtc.ts
```diff
@@ -613,7 +614,11 @@ export class WebrtcExtension extends EventEmitter implements StreamingClientExte
    * Stanza Handlers
    */
   private async handlePropose (msg: ProposeStanza) {
-    if (msg.from === this.jid) {
+    // if (msg.from === this.jid) {
+    //   return;
+    // }
+
+    if (msg.from in this.sessionsMap) {
```
hjon, line 621: I might be missing something about what you were thinking with this check, but I think we should just using the existing check. `sessionsMap` is storing session IDs, not JIDs, right?

maxwellmooney13: Yeah but this sorta falls into my confusion before... I _thought_ this was an appropriate replacement for commenting out the block that I did but I'm more and more convinced we don't need to comment out that block to begin with

src/webrtc.ts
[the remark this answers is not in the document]
zservies: Jon is right, the session IDs will be the same.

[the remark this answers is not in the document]
zservies: I do wonder if there's any actual need to create a map. Since we know that screen recording won't use SDP over XMPP, i think we can could just create a helper function to set this value.

zservies (changes requested):
Let's type the map.

test/unit/webrtc.test.ts
```diff
@@ -666,7 +667,7 @@ describe('handlePropose', () => {
     expect(spy).toHaveBeenCalled();
   });
 
-  it('should track sdpOverXmpp', async () => {
+  it.skip('should track sdpOverXmpp', async () => {
```
zservies, line 670: Remove skip.

test/unit/session-initiation-timing.test.ts
```diff
@@ -104,7 +104,7 @@ describe('handling of Jingle session-initiate and SDP offer with different order
     webrtc.pendingSessions['session24'] = pendingSession;
   });
 
-  it('should only create a GenesysCloudMediaSession and set the meetingId if SDP is handled first', async () => {
+  it.skip('should only create a GenesysCloudMediaSession and set the meetingId if SDP is handled first', async () => {
```
zservies, line 107: Remove.

zservies (approved):

PR #299, opened 2025-05-23
[STREAM-63] - Added in check to assign customHeaders from config if none passed in

maxwellmooney13:


src/client.ts
```diff
@@ -75,7 +75,7 @@ export class Client extends EventEmitter {
 
   constructor (options: IClientOptions) {
     super();
-    this.http = new HttpClient();
```
maxwellmooney13, line 78: When Streaming Client spins up a `HttpClient` instance, it now passes in the customHeaders from the config (though perhaps we should check if it exists first?) so that the `HttpClient` can then use it at a later point

src/http-client.ts
```diff
@@ -67,10 +74,15 @@ export class HttpClient {
     const logger = opts.logger || console;
     const start = new Date().getTime();
 
+    opts.customHeaders = {
+      ...this.customHeaders,
+      ...opts.customHeaders
+    };
```
maxwellmooney13, line 80: Oh this needs a change. I meant for this to include `|| {}` in the case of either of these being undefined or something like that.

Otherwise the thinking here (which Jon and I worked through) is that though we can't think of a reason why this would be the case, but if there is a scenario where we need to pass in a different value for `customHeaders` than what is passed in for the Streaming Client config, then we should use that passed in value _over_ the config value which is hopefully what will happen here.  Again, we can't think of a reason this would happen, but if someone passes in a different `customHeaders`, there would be a _reason_ behind it so it needs to be honored.

src/client.ts
```diff
@@ -671,8 +670,7 @@ export class Client extends EventEmitter {
         method: 'post',
         host: this.config.apiHost,
         authToken: this.config.authToken,
-        logger: this.logger,
-        customHeaders: this.config.customHeaders
```
maxwellmooney13, line 675: Removed these because they get added in regardless in `requestApi`, so it seemed a bit redundant for the same `customHeaders`

hjon (approved):

zservies (approved):

PR #306, opened 2025-07-09
[STREAM-154] Track StanzaMediaSessions across disconnects

hjon:


src/webrtc.ts
```diff
@@ -394,7 +406,6 @@ export class WebrtcExtension extends EventEmitter implements StreamingClientExte
     const pendingSession = this.pendingSessions[options.sid!];
     if (pendingSession) {
       delete this.pendingSessions[pendingSession.sessionId];
-      delete this.sessionsMap[pendingSession.sessionId];
```
hjon, line 397: I changed this to delete from the `sessionsMap` when the session terminates (line 434), which also matches how we do it with GenesysCloudMediaSessions.

zservies (approved):

PR #308, opened 2025-07-10
[STREAM-653] - Added fields to upgradeMediaPresence stanza definition.

zservies:


maxwellmooney13 (approved):

zservies:
I think we're going to need to revert the webpack bump, there are a lot of dependencies that aren't compatible with it and its causing a lot of problems in the build process.

PR #314, opened 2025-07-22
Release/v19.3.1

zservies:


maxwellmooney13 (approved):

zservies:
Had some conflicts to sort through which is why the history is a little borked. This just contains an axios + logger dependency update to fix a critical Snyk vulnerability. I also added a new GH action step to run npm audit so we can catch that earlier,.

RobinsonWM (approved):

PR #316, opened 2025-07-29
Stream-643: bump webpack to 5.94.0

AidanWZ:


tslint.json
```diff
@@ -7,6 +7,7 @@
     ]
   },
   "rules": {
+    "deprecation": false,
```
zservies: I think I know the answer, but is it a simple task to migrate from TSLint to ESLint? TSLint is technically deprecated itself. When I explored updating I figured we'd need to do that at the same time.

zservies (changes requested):
I think I know the answer, but is it a simple task to migrate from TSLint to ESLint? TSLint is technically deprecated itself. When I explored updating I figured we'd need to do that at the same time.

tslint.json
AidanWZ: migrated to ESLint ✅

hjon (approved):

zservies (approved):

PR #318, opened 2025-09-09
STREAM-892: add enablePartialBulkResubscribe option for notifications

jasonmobley:
Adds a new optional config property `enablePartialBulkResubscribe` that modifies the behavior of the Notifications module bulk subscribe to include `?ignoreErrors=true` on the PUT request to update the subscription. The bulk subscribe method is updated to always return a Promise for a BulkSubscribeResult that contains the status of each topic. This behavior is opt-in, so the default behavior stays the same as before.

The `?ignoreErrors=true` param on the public API endpoint corresponds to the `nofail` mode in hawk-consumer which makes bulk requests succeed with individual failure results instead of failing the entire bulk operation due to one of the topics failing (e.g. due to lack of permissions).

### DIFF
--- a/src/notifications.ts
+++ b/src/notifications.ts
@@ -1,11 +1,13 @@
 import { Agent } from 'stanza';
 import { PubsubEvent, PubsubSubscription, PubsubSubscriptionWithOptions } from 'stanza/protocol';
-const debounce = require('debounce-promise');
+import debounce from 'debounce-promise';
 
 import { Client } from './client';
-import { RequestApiOptions, StreamingClientExtension } from './types/interfaces';
+import { IClientOptions, RequestApiOptions, StreamingClientExtension } from './types/interfaces';
 import { NamedAgent } from './types/named-agent';
 import { splitIntoIndividualTopics } from './utils';
+import { StreamingSubscriptionError } from './';
+import { AxiosResponse } from 'axios';
 
 const PUBSUB_HOST_DEFAULT = 'notifications.mypurecloud.com';
 const MAX_SUBSCRIBABLE_TOPICS = 1000;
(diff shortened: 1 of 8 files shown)


maxwellmooney13 (approved):

test/unit/notifications.test.ts
```diff
@@ -109,6 +109,111 @@ describe('Notifications', () => {
...
+        await Promise.all([
+          notification.subscribe('a.topic').then(() => { aResult = 'fulfilled'; }, () => { aResult = 'rejected'; }),
+          notification.subscribe('b.topic').then(() => { bResult = 'fulfilled'; }, () => { bResult = 'rejected'; }),
+          notification.subscribe('c.topic').then(() => { cResult = 'fulfilled'; }, () => { cResult = 'rejected'; }),
+        ]);
+        expect(aResult).toEqual('fulfilled');
+        // b.topic subscribe() call is rejected because bulk subscribe result is "Rejected" for that topic
+        expect(bResult).toEqual('rejected');
+        // c.topic subscribe() call is rejected because bulk subscribe result is "Unknown" for that topic (not in the API response)
+        expect(cResult).toEqual('rejected');
```
hjon, line 223: It looks like with `ignoreErrors=true` and a topic that will be rejected (e.g. `b.topic`), the subscription attempt runs through the topic that will be rejected (e.g. attempt to subscribe to `a.topic`, attempt to subscribe to `b.topic`), then stops and the topics **after** the rejected topic aren't tried (e.g. the attempt to subscribe `c.topic` doesn't happen). Does that match what you've seen from the backend?

hjon (approved):

test/unit/notifications.test.ts
jasonmobley: Nah, the backend behavior is what you'd expect: all topics get processed regardless of order or other failures.

The stuff for `c.topic` here is really just to achieve 100% branch coverage. There's logic around if the API response doesn't include a topic we were expecting a result for and that's the case being tested by `c.topic`. I could spell that out in more code comments on the mocks above if you think it needs clarification.

hjon: Ah! I was thrown off, so I wouldn't mind a little clarification in the code, but I'm also pretty satisfied with this being our record of the behavior. Or maybe add one more topic in the mock so we can show both aspects (all topics as processed **and** how we handle a topic not being included in the results)?

jasonmobley: See commit 445d4fd6 for clarified test behavior with more comments and test topics with more descriptive names... `a.ok.topic`, `b.bad.topic`, `c.unknown.topic`

hjon:
@jasonmobley Could you update the changelog? (Sorry I didn't catch that earlier.)

src/notifications.ts
```diff
@@ -22,15 +24,19 @@ export class Notifications implements StreamingClientExtension {
   subscriptions: any;
   bulkSubscriptions: any;
   topicPriorities: any;
-  debouncedResubscribe: any;
+  debouncedResubscribe: () => Promise<BulkSubscribeResult>;
```
zservies, line 27: Thank you for leaving things better than you found them <3

zservies:
Looks good to me, I'll stamp and merge once the changelog is updated.

zservies (approved):

PR #323, opened 2025-09-29
[STREAM-941] fix notifications.subscribe() result for internally combined topics

jasonmobley:
Fixes a couple issues related to the internal topic combining that caused incorrect or unexpected results for individual topic subscribes.

Example:
```ts
subscribe("topic.ok"); // expected to resolve
subscribe("topic.bad"); // expected to reject
```
Internally, this becomes a single entry in the bulk resubscribe: `topic?ok&bad`
Before this change, the code that looked for a result only knew to look for for `topic.ok` and `topic.bad` individually. It did not normalize the actual response entity `topic?ok&bad` to produce the individual results. Similarly, the bulk subscribe result generation was only using topics from the requested list to generate the result rather than starting from everything in the API response and only supplementing as needed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,9 +42,9 @@
         "@babel/runtime": "^7.12.0",
         "@rollup/plugin-commonjs": "^22.0.0-1",
         "@rollup/plugin-node-resolve": "^13.0.6",
+        "@types/debounce-promise": "^3.1.9",
         "@types/eslint": "^8.56.12",
         "@types/estree": "^1.0.8",
-        "@types/debounce-promise": "^3.1.9",
         "@types/jest": "^26.0.3",
         "@types/lodash.throttle": "^4.1.6",
         "@types/nock": "^11.1.0",
@@ -3330,6 +3330,12 @@
         "node": ">=6.9.0"
       }
     },
+    "node_modules/@types/debounce-promise": {
+      "version": "3.1.9",
+      "resolved": "https://registry.npmjs.org/@types/debounce-promise/-/debounce-promise-3.1.9.tgz",
+      "integrity": "sha512-awNxydYSU+E2vL7EiOAMtiSOfL5gUM5X4YSE2A92qpxDJQ/rXz6oMPYBFDcDywlUmvIDI6zsqgq17cGm5CITQw==",
+      "dev": true
+    },
     "node_modules/@types/eslint": {
       "version": "8.56.12",
       "resolved": "https://registry.npmjs.org/@types/eslint/-/eslint-8.56.12.tgz",
@@ -3350,12 +3356,6 @@
         "@types/eslint": "*",
         "@types/estree": "*"
       }
-    },
-    "node_modules/@types/debounce-promise": {
-      "version": "3.1.9",
-      "resolved": "https://registry.npmjs.org/@types/debounce-promise/-/debounce-promise-3.1.9.tgz",
-      "integrity": "sha512-awNxydYSU+E2vL7EiOAMtiSOfL5gUM5X4YSE2A92qpxDJQ/rXz6oMPYBFDcDywlUmvIDI6zsqgq17cGm5CITQw==",
-      "dev": true
     },
     "node_modules/@types/estree": {
       "version": "1.0.8",
--- a/src/notifications.ts
+++ b/src/notifications.ts
@@ -411,11 +411,26 @@ export class Notifications implements StreamingClientExtension {
     if (response && response.data && 'entities' in response.data && Array.isArray(response.data.entities)) {
       topicResponseEntities = response.data.entities;
     }
-    const topicResponsesById: { [topic: string]: ChannelTopicResponseEntity } = {};
-    for (const topicEntity of topicResponseEntities) {
-      topicResponsesById[topicEntity.id] = topicEntity;
-    }
     const result: BulkSubscribeResult = {};
+    for (const topicEntity of topicResponseEntities) {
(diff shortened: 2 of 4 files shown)


package-lock.json
jasonmobley, line 1: Not sure if this was wrong from my last ticket, but the lockfile updated when I ran npm install so I committed that.

src/notifications.ts
```diff
@@ -411,31 +411,42 @@ export class Notifications implements StreamingClientExtension {
...
-      if (this.enablePartialBulkResubscribe) {
-        if (topic in topicResponsesById) {
-          const { state, rejectionReason } = topicResponsesById[topic];
-          result[topic] = { topic, state, rejectionReason };
-        } else {
-          result[topic] = { topic, state: 'Unknown' };
-        }
-      } else {
-        result[topic] = { topic, state: 'Permitted' };
-      }
```
jasonmobley, line 436: Part of why this was wrong is it was only looping through the bulk topics list `topics` rather than through all the requested topics which is what `toSubscribe` is. The other piece was to use the API response as more of the "source of truth" for the bulk result and then only add in stuff from `toSubscribe` below if it wasn't in the bulk response.

package-lock.json
zservies: No this should be good. Your last lockfile commit originally had some nexus stuff in it because you probably had an npmrc in your project or under your user that it was trying to use. Since this is just open-source, the registry we pull everything is just the default npmjs registry.

jasonmobley:
Checking in on the status of this PR... anything else needed from me? @zservies @hjon

test/unit/notifications.test.ts
```diff
@@ -226,6 +236,30 @@ describe('Notifications', () => {
...
+        expect(bResult).toEqual('rejected');
+        expect(Object.keys(notification.bulkSubscriptions)).toContain('topic.ok');
+        expect(Object.keys(notification.bulkSubscriptions)).toContain('topic.bad');
+      });
+      // Not sure this is desired behavior, but it's the actual behavior so this test documents that... :)
+      it('should succeed subscribe() call for combined topic even if individual topic fails', async () => {
+        let aResult: undefined | 'fulfilled' | 'rejected';
+        // In this case the API response is expected to have a "Permitted" result for "topic?ok&bad"
+        // But will also have a "Rejected" result for individual "topic.bad" (which we would only
+        // know if we had explicitly subscribed that topic).
```
hjon, line 258: Just to make sure I'm understanding this correctly...

If we `subscribe` to `topic?ok&bad`, the API response will tell us that `topic?ok&bad` succeeded/is `Permitted`, but that `topic.bad` failed/is `Rejected`. And in `bulkSubscribe`, it looks like the result we return will reflect that. Something like `{ topic.bad: { topic: 'topic.bad', state: 'Rejected' }, topic?ok&bad: { topic: 'topic?ok&bad', state: 'Permitted' } }`? But when `subscribe` looks at that, it will only look at the result for the topic passed in (`topic?ok&bad`) and therefore it will look like it succeeded overall.

Am I tracking correctly?

jasonmobley: Yes, you're understanding is correct. This only applies if the **requested** topic was already a combined identifier.

Expectations are:
- subscribe("topic?ok&bad") will resolve as successful even though topic.bad actually failed
- subscribe("topic.ok") on its own will resolve success (even if it ends up combined in the bulk resub)
- subscribe("topic.bad") on its own will reject as failure (even if it ends up combined in the bulk resub)

jasonmobley:
Bumping this one @hjon

src/notifications.ts
```diff
@@ -411,31 +411,42 @@ export class Notifications implements StreamingClientExtension {
     if (response && response.data && 'entities' in response.data && Array.isArray(response.data.entities)) {
       topicResponseEntities = response.data.entities;
     }
-    const topicResponsesById: { [topic: string]: ChannelTopicResponseEntity } = {};
+    const result: BulkSubscribeResult = {};
     for (const topicEntity of topicResponseEntities) {
-      topicResponsesById[topicEntity.id] = topicEntity;
+      const { id, state, rejectionReason } = topicEntity;
+      const topicResult = result[id] = { topic: id, state, rejectionReason };
```
zservies: NIT: I personally prefer this broken up for readability.

src/notifications.ts
```diff
@@ -411,31 +411,42 @@ export class Notifications implements StreamingClientExtension {
...
+      // If response entity is a combined topic ID like "a.b?c&d" include individualized topic IDs
+      // as keys in the map. This could either point to the same result as the combined topic ID
+      // or to a specific result for that individual topic if backend provides a specific result.
+      // Example: caller asked to subscribe "a.b?c&d" but user lacks permission for topic "a.b.d"
+      // In this case, API response will include "a.b?c&d" as success along with "a.b.d" as failure.
+      if (id.includes('?')) {
+        for (const individualTopic of splitIntoIndividualTopics(id)) {
+          const hasIndividualTopicResult = (individualTopic in result);
```
zservies: NIT: It might be safer to use `Object.prototype.hasOwnProperty` or `Object.hasOwn`.

zservies (approved):

src/notifications.ts
jasonmobley: Addressed in 716622bb

src/notifications.ts
jasonmobley: Addressed in 716622bb

hjon (approved):

PR #325, opened 2025-11-10
[STREAM-978] - Cleanup after failed ICE state.

zservies:


zservies:
I think we might already be cleaning up interruptions in `genesys-cloud-media-session.ts` https://github.com/purecloudlabs/genesys-cloud-streaming-client/blob/develop/src/types/genesys-cloud-media-session.ts#L227.

PR #326, opened 2025-11-10
[STREAM-978] - Cleanup after failed ICE state.

zservies:


zservies:
I think we might already be cleaning up interruptions in genesys-cloud-media-session.ts https://github.com/purecloudlabs/genesys-cloud-streaming-client/blob/develop/src/types/genesys-cloud-media-session.ts#L227.

patrickfeeney03 (approved):

maxwellmooney13 (approved):

src/types/stanza-media-session.ts
```diff
@@ -202,6 +202,8 @@ export class StanzaMediaSession extends MediaSession implements IMediaSession {
         candidatesDiscovered: this.iceCandidatesDiscovered,
         candidatesReceivedFromPeer: this.iceCandidatesReceivedFromPeer
       });
+      // Cleanup the session because failed is a state of finality.
+      this.end('failed-transport');
```
hjon, line 206: We shouldn't have to do this, because stanza already takes care of that [in `ICESession`](https://github.com/legastero/stanza/blob/master/src/jingle/ICESession.ts#L331) and we ultimately extend that.

hjon:
> I think we might already be cleaning up interruptions in genesys-cloud-media-session.ts https://github.com/purecloudlabs/genesys-cloud-streaming-client/blob/develop/src/types/genesys-cloud-media-session.ts#L227.

I'll admit that I still don't feel super comfortable about the difference between `connectionState` and `iceState`, but it *might* still be worth checking when the `iceState` changes. I linked to this in the other comment, too, but stanza uses the combination of the two states to decide whether to `end` or just restart the ICE process: https://github.com/legastero/stanza/blob/master/src/jingle/ICESession.ts#L331

PR #327, opened 2025-11-11
STREAM-824: add new session type handle live monitoring sdps

AidanWZ:


src/types/interfaces.ts
```diff
@@ -124,13 +124,14 @@ export interface IError {
   stack?: string;
 }
 
-export type SessionTypesAsStrings = 'softphone' | 'screenShare' | 'screenRecording' | 'collaborateVideo' | 'unknown';
+export type SessionTypesAsStrings = 'softphone' | 'screenShare' | 'screenRecording' | 'liveScreenMonitoring' | 'collaborateVideo' | 'unknown';
```
AidanWZ, line 127: lmk what people think about the session type name, could also be just `liveMonitoring`

bruegg (approved):

maxwellmooney13 (approved):

src/types/interfaces.ts
zservies: I'm good with either. I'm not sure it'll ever be expanded beyond screen-monitoring so I think this is good.

src/webrtc.ts
```diff
@@ -1049,6 +1049,8 @@ export class WebrtcExtension extends EventEmitter implements StreamingClientExte
       return SessionTypes.acdScreenShare;
     } else if (isScreenRecordingJid(jid)) {
       return SessionTypes.screenRecording;
+    } else if (isLiveScreenMonitoringJid(jid)) {
+      return SessionTypes.liveScreenMonitoring;
```
zservies, line 1053: I think we already do, but double check that we have access to the `roomJid` in the offer. We might not need to do anything extra to extract it, but Volt will be depending on it for routing to the right place.

zservies:
Don't forget to update the changelog.

hjon (approved):

src/webrtc.ts
AidanWZ: Looking at media-controller code the `from` field on the IQ is necessarily populated, so translation to sessionType should be 👍

zservies (approved):

PR #330, opened 2025-11-19
[STREAM-997] - Retry API requests on network failures. Added max retry attempts.

zservies:


src/http-client.ts
```diff
@@ -29,14 +29,30 @@ export class HttpClient {
...
+    'ECONNABORTED',
+    'ERR_NETWORK',
+    'ETIMEDOUT'
+  ]);
+
   constructor (httpClientOptions?: IHttpClientOptions) {
     this.customHeaders = httpClientOptions?.customHeaders;
   }
 
   requestApiWithRetry<T = any> (path: string, opts: RequestApiOptions, retryInterval?: number): RetryPromise<T> {
```
zservies, line 42: I think we should honestly improve this to add real exponential backoff but I don't want to do that right now at risk of blowing something up. Maybe we should just expose a new function, `requestApiWithBackoff` that can be opted into.

src/http-client.ts
```diff
@@ -53,6 +69,15 @@ export class HttpClient {
           }
         }
 
+        // Check network error codes independently - error may have both response (with status 0) and a network error code.
+        if (error?.code && retryValue === false) {
```
zservies, line 73: We're checking for `=== false` instead of `!retryValue` because its possible to get a `retryValue = 0` scenario where we are supposed to retry immediately.

src/http-client.ts
```diff
@@ -29,14 +29,30 @@ export class HttpClient {
     504
   ]);
 
+  static retryNetworkErrorCodes = new Set([
```
zservies, line 32: There may be more network codes that we should retry on. These seem like the obvious three.

zservies:
Basically this change is supposed to add the ability to retry when our request doesn't make it to the server or we just receive no response back. The max attempt stuff I threw in because it didn't make much sense to just hammer the server indefinitely.

maxwellmooney13 (approved):

hjon (approved):

PR #332, opened 2025-11-27
[Stream-995] - Add 'missingPermissions' property to StreamingSubscriptionError

patrickfeeney03:
https://inindca.atlassian.net/browse/STREAM-995

Hawk Consumer now includes a new property which shows which permissions a user is missing when trying to subscribe to a topic.

This is from trying to subscribe to something we are not allowed to in WebDir ran locally. The error now contains the `missingPermissions` property.
image

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -27,9 +27,11 @@ export class StreamingClientError extends Error {
 
 export class StreamingSubscriptionError extends Error {
   name = 'StreamingSubscriptionError';
+  missingPermissions?: string[];
 
-  constructor (message: string, public readonly topic?: string, public readonly operation?: 'subscribe' | 'unsubscribe') {
+  constructor (message: string, public readonly topic?: string, public readonly operation?: 'subscribe' | 'unsubscribe', missingPermissions?: string[]) {
     super(message);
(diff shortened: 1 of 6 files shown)


AidanWZ (approved):

zservies (approved):

maxwellmooney13 (approved):

hjon (approved):
Changelog. 🙂

src/utils.ts
```diff
@@ -27,9 +27,11 @@ export class StreamingClientError extends Error {
 
 export class StreamingSubscriptionError extends Error {
   name = 'StreamingSubscriptionError';
+  missingPermissions?: string[];
 
-  constructor (message: string, public readonly topic?: string, public readonly operation?: 'subscribe' | 'unsubscribe') {
+  constructor (message: string, public readonly topic?: string, public readonly operation?: 'subscribe' | 'unsubscribe', missingPermissions?: string[]) {
```
jasonmobley: This is fine, but you might consider making this new param an optional object called `details` of which `missingPermissions` is its only field for now. That way we don't have more and more positional params in the future if other bits of detail get added later.

jasonmobley (approved):

PR #335, opened 2025-12-16
[STREAM-949] - Catch and emit error in `handleStanzaDisconnectedEvent`

patrickfeeney03:
If `autoReconnect` was true and for some reason the call to `connect` failed, an uncatchable error would be thrown.

The Journey Management UI team asked us to handle this error to prevent their app from breaking since they can't catch the error.

The error is now caught and emitted under 'disconnected'.


test/scripts/token-becoming-invalid.ts
```diff
@@ -0,0 +1,77 @@
+interface Window {
```
patrickfeeney03: This file is for local testing. Ignore

maxwellmooney13 (approved):

src/client.ts
```diff
@@ -259,7 +259,8 @@ export class Client extends EventEmitter {
     this.emit('disconnected', { reconnecting: this.autoReconnect });
 
     if (this.autoReconnect) {
-      return this.connect({ keepTryingOnFailure: true });
+      return this.connect({ keepTryingOnFailure: true })
+        .catch(e => this.emit('error', e));
```
hjon: I'm still pondering this a bit, but here's my current thinking.

Initially, I thought that we should just catch the error without emitting because this is an automatic reconnect. But I also think it's important for the consuming code to be able to know what's happening, especially in the reported scenario where it's an auth issue.

I believe this is the sequence in this scenario:
- A `disconnected` event is emitted
- We try to connect
- That fails, and this error is emitted or thrown

This is the point where I'm still a little torn on the best way forward.

One part of me leans towards changing this to emit a `disconnected` event again, but include an error. That makes it a little easier to associate the error with the connection attempt. And since we already emitted a `disconnected` event with `reconnecting: true`, I like the idea of an additional `disconnected` event to show that the reconnect failed. I mostly land here, but I think I could be convinced otherwise.

The other part of me says to just leave it as an `error` because this isn't associated with the original connection request, but one that we performed automatically.

In either case, I think it would be a good idea for the error to be a `StreamingClientError` and associate it with the right type (e.g. in the reported scenario it should be `StreamingClientErrorTypes.invalid_token`).

I have a strong belief that errors from libraries should be consistent, even if their dependencies change. So in this case, the error should not be from Axios because, if we ever change that library, the error could change and be a breaking change. I don't mind passing along the AxiosError for more information, but I expect the `StreamingClientError` should have sufficient information for consuming code to make any decisions (e.g. get the user to login again).

patrickfeeney03: Talked with Jon about this.

index.html
```diff
@@ -29,6 +29,7 @@
     // loadScript('test/scripts/happy-path.js');
     // loadScript('test/scripts/duplicate-channel-connection.js');
     // loadScript('test/scripts/send-message.js');
+    loadScript('test/scripts/token-becoming-invalid.js');
```
zservies: Should this be commented out?

src/client.ts
```diff
@@ -259,7 +259,18 @@ export class Client extends EventEmitter {
     this.emit('disconnected', { reconnecting: this.autoReconnect });
 
     if (this.autoReconnect) {
-      return this.connect({ keepTryingOnFailure: true });
+      return this.connect({ keepTryingOnFailure: true })
+        .catch(error => {
+          this.logger.error('Failed autoReconnect', {
```
zservies: Should we change this error to `Failed to auto reconnect`?

zservies (approved):

src/client.ts
hjon: To expand on this:
1. I overlooked that this is already going to be a `StreamingClientError` because that's what we throw from `connect`. So we're good there.
2. Patrick asked about how to let the consuming code know that this instance of streaming-client was "done" - it had disconnected and wasn't going to try reconnect again. From our discussion, we landed on passing the error and `reconnecting: false` (I had forgotten about that).

hjon (approved):

PR #338, opened 2026-02-12
[STREAM-1179] Allow a streaming-client instance to be marked as alertable

hjon:


maxwellmooney13 (approved):

AidanWZ (approved):

src/types/interfaces.ts
```diff
@@ -25,17 +25,24 @@ export interface IClientOptions {
...
   enablePartialBulkResubscribe?: boolean;
+  /** Genesys internal use only - non-Genesys apps that pass in `alertableInteractions` may experience unexpected behavior */
+  alertableInteractions?: IAlertableInteractions;
 }
 
 export interface ICustomHeader {
   [header: string]: string;
 }
 
+export interface IAlertableInteractions extends Object {
```
zservies: We shouldn't need to extend `Object`, it should be implicitly extended already.

src/types/interfaces.ts
```diff
@@ -25,17 +25,24 @@ export interface IClientOptions {
...
+  voice?: boolean;
```
zservies: In the future will we need to add `video`, `digital`, etc?

src/types/interfaces.ts
zservies: Should this be a regular object or an array? I'm not sure I have a particularly strong opinion, but it did initially seem strange for this to be an object.

I figured we'd do this somewhat like we do `allowedSessionTypes` in the SDK - so it'd be something like `['voice', 'digital']` in the future. I suppose this lets us get even more specific within the space of `digital` but even then we could always make this an array of objects.

src/alerting-leader.ts
```diff
@@ -0,0 +1,66 @@
...
+      logger: this.client.logger,
+      data: {
+        alertable: true
+      }
+    };
+
+    // STREAM-1204
+    // There's a race condition between the backend service knowing about the connection
+    // and us marking the connection as alertable. For now, we'll just retry with some delay.
+    const maxRetries = 16;
```
zservies, line 37: My head says this is a big number but its only eight seconds of retries.

src/client.ts
```diff
@@ -686,19 +690,24 @@ export class Client extends EventEmitter {
     }
 
     if (this.hardReconnectRequired) {
-      let jidPromise: Promise<any>;
+      let userPromise: Promise<any>;
 
-      if (this.config.jid) {
-        jidPromise = Promise.resolve(this.config.jid);
+      if (this.config.userId && this.config.jid) {
```
zservies, line 695: Should this be an `&&`? The `userId` field is optional it looks.

src/alerting-leader.ts
hjon: Yeah...of course, "only" eight seconds is rather long...but I'm just trying to avoid any issues until we can improve it (I hope).

src/types/interfaces.ts
hjon: I'll think about that. I don't quite remember why I landed here, but the reasons may no longer be relevant.

src/types/interfaces.ts
hjon: Yep. So to your earlier comment, an array might be just as good.

src/client.ts
hjon: Both `userId` and `jid` are optional. But we need the `userId` for alerting leader, so I didn't want to resolve the promise from the config if we didn't have both.

src/types/interfaces.ts
hjon: I took another look and I think I agree with you. I suspect I had a certain idea of how this would be used that didn't pan out.

(I think Max made a similar suggestion early on, so between the two of you I came around. 🙂)

zservies (approved):

PR #340, opened 2026-03-03
[STREAM-1271] - catch debouncedResubscribe error

patrickfeeney03:
Somebody from the JM team was asking if we could catch this.
`handleStanzaInstanceChange` calls `debouncedResubscribe` and does not wait for it or attaches any handlers. So if `debouncedResubscribe` fails, nothing catches it, and it bubbles up to the GlobalHandler.

### DIFF
--- a/src/notifications.ts
+++ b/src/notifications.ts
@@ -63,7 +63,11 @@ export class Notifications implements StreamingClientExtension {
 
     if (needsToResub) {
       this.client.logger.info('resubscribing due to hard reconnect');
-      void this.debouncedResubscribe();
+      this.debouncedResubscribe().catch((err) => {
+        const msg = 'Error resubscribing to topics';
+        this.client.logger.error(msg, err);
+        this.client.emit('pubsub:error' as any, { msg, err });
+      });
     }
   }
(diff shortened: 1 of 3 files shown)


src/notifications.ts
```diff
@@ -63,7 +63,11 @@ export class Notifications implements StreamingClientExtension {
 
     if (needsToResub) {
       this.client.logger.info('resubscribing due to hard reconnect');
-      void this.debouncedResubscribe();
+      this.debouncedResubscribe().catch((err) => {
```
patrickfeeney03, line 66: This is the same we do in line 304, in `subscriptionsKeepAlive`

AidanWZ (approved):

src/notifications.ts
patrickfeeney03, line 66: Since `debouncedResubscribe` is async and we were not awaiting it or attaching any sort of handler, it was thrown into the void, and if it rejected while making a request to pub api the client couldn't do anything about that.

hjon (approved):

zservies (approved):

PR #341, opened 2026-03-16
STREAM-838: add more robust internet connectivity check

AidanWZ:


src/client.ts
```diff
@@ -596,13 +595,50 @@ export class Client extends EventEmitter {
...
+      this.logger.warn('navigator.onLine reports offline — connectivity may be unavailable');
+      this.emit('networkConnectivityWarning', { reason: 'navigator.onLine is false' });
+      return false;
+    }
+
+    try {
+      const opts: RequestApiOptions = {
+        method: 'get',
+        host: this.config.apiHost,
+        authToken: this.config.authToken || this.config.jwt,
```
zservies: Will this work with a JWT?

src/client.ts
```diff
@@ -596,13 +595,50 @@ export class Client extends EventEmitter {
...
+        logger: this.logger,
+        requestTimeout: 10000
+      };
+      await this.http.requestApi('users/me', opts);
+      return true;
+    } catch (err) {
+      this.logger.warn('Active network connectivity check failed — connectivity may be unavailable', { error: err });
+      this.emit('networkConnectivityWarning', { reason: 'active connectivity check failed', error: err });
```
zservies, line 633: Is this a new event?

src/client.ts
AidanWZ: nope, just saw the examples in the same file, removed that change. Instead i will only perform the api based check for auth token clients

src/client.ts
```diff
@@ -596,13 +595,57 @@ export class Client extends EventEmitter {
...
     if (this.cancelConnectionAttempt) {
       throw new UserCancelledError('Connection attempt cancelled');
     }
 
-    if (!navigator.onLine) {
-      throw new OfflineError('Browser is offline, skipping connection attempt');
+    // navigator.onLine is unreliable — use it as a hint, not a gate.
+    // Run an active connectivity check to verify real network access.
+    // Either way, we still proceed with the connection attempt.
+    const isConnected = await this.checkNetworkConnectivity();
```
hjon: Since we don't want to gate connection attempts, I don't think we should `await` for this, either (especially with the current request timeout of 10s).

Changing the request timeout to be shorter would help, but even then I would prefer to let a connection attempt happen as fast as possible.

AidanWZ: good call, will make the call fire and forget

src/client.ts
```diff
@@ -596,14 +595,59 @@ export class Client extends EventEmitter {
...
+      this.logger.warn('navigator.onLine reports offline — connectivity may be unavailable');
+      this.emit('networkConnectivityWarning', { reason: 'navigator.onLine is false' });
+      return false;
+    }
+
+    // JWT-based connections (e.g. background assistants) don't have an auth token
+    // that works with the users/me endpoint, so we can only rely on navigator.onLine
+    if (this.config.jwt && !this.config.authToken) {
+      this.logger.debug('Skipping active connectivity check in JWT mode, relying on navigator.onLine');
+      return true;
```
patrickfeeney03, line 618: Is there no other endpoint available when using JWT? Does the endpoint **need** to be part of genesys?
Could we try publics api health endpoint: `https://api.inindca.com/api/v2/health/check` thats dca it'd be different for each region
and if for some reason thats not reachable we could just try to reach google.com or something idk

patrickfeeney03 (changes requested):

src/client.ts
AidanWZ: yes, updated the documentation

src/client.ts
AidanWZ: opting to just skip a connectivity check for JWT users, i verified that a user cannot access the GC api with a JWT and we only really care if they can access our apis anyway

zservies (approved):

hjon (approved):

PR #343, opened 2026-04-14
[STREAM-1155] Allow clients to become the alerting leader and listen for alerting leader events

hjon:


src/types/interfaces.ts
```diff
@@ -37,6 +37,16 @@ export enum AlertableInteractionTypes {
   voice = 'voice'
 }
 
+export interface IAlertingStatus {
+  alerting: boolean;
+  configured: boolean;
+  clientType?: string;
+}
+
+export interface ILeaderStatus {
```
zservies, line 61: Can you add a comment to specify this is the actual alerting leader's status? I might be slow but for a second I didn't understand what `ILeaderStatus` is.

src/alerting-leader.ts
```diff
@@ -1,24 +1,61 @@
...
     }
   }
 
+  private async subscribeToAlertingLeader (): Promise<any> {
+    const topic = `v2.users.${this.client.config.userId}.alertingleader`;
+    this.client.on(`notify:${topic}`, (event) => {
+      this.abortController?.abort();
+
+      if (event.eventBody?.connectionId) {
+        const alerting = event.eventBody.connectionId === this.connectionId;
```
zservies: This is basically a check for "am I the alerting leader" right?

src/types/interfaces.ts
```diff
@@ -37,6 +37,16 @@ export enum AlertableInteractionTypes {
   voice = 'voice'
 }
 
+export interface IAlertingStatus {
+  alerting: boolean;
```
zservies, line 48: I essentially asked this in another comment, but this is basically a boolean that is `true/false` if we are the alerting leader?

src/alerting-leader.ts
```diff
@@ -56,11 +93,73 @@ export class AlertingLeaderExtension implements StreamingClientExtension {
...
+      method: 'get',
+      host: this.client.config.apiHost,
+      authToken: this.client.config.authToken,
+      logger: this.client.logger,
+      signal: this.abortController.signal
+    };
+
+    try {
+      const currentLeader = await this.client.http.requestApi('users/alertingleader', leaderRequestOptions);
+      const alerting = currentLeader.data.connectionId === this.connectionId;
```
zservies: Just answered my own questions. Its probably worth putting a couple comments explaining what `alerting` is in the interface.

zservies: NIT: I guess we could also explicitly mark these as `boolean` too.

src/types/interfaces.ts
```diff
@@ -37,6 +37,16 @@ export enum AlertableInteractionTypes {
...
+  configured: boolean;
+  clientType?: string;
```
zservies, line 57: Should we include a list of client types as an enum?

src/alerting-leader.ts
```diff
@@ -1,24 +1,61 @@
...
+  }
+
+  private async setupAlertingLeader () {
     if (this.alertableInteractionTypes.length !== 0) {
-      this.markAsAlertable();
+      try {
+        await this.subscribeToAlertingLeader();
+        await this.markAsAlertable();
+        await this.getAlertingLeader();
+      } catch (err) {
```
zservies, line 32: I'm not sure if its useful information, but we could specify which request failed (subscription, marking, fetching).

src/types/interfaces.ts
hjon: Mostly. It's trying to encompass the general idea of "should I be alerting?", which can be because this client is the alerting leader or because alerting leader is enabled but not configured by an admin or because the setup failed and we're falling back to alerting status.

From discussions @maxwellmooney13, we landed on it being helpful for consumers to have a single property to check rather than something like `alerting || !configured` in multiple places. Looking at that now, that might not be too bad; I don't remember if we thought it would be more involved at some point.

src/types/interfaces.ts
hjon: I think that's a good idea, but I'm probably going to kick it down the road. Even though I tried to anticipate future use cases a bit, I feel uncertain about this one, maybe because of the potential for custom clients someday?

src/alerting-leader.ts
hjon: I don't think I'll worry about the specifics at this point, but it does seem worth adding a log.

src/alerting-leader.ts
hjon: Correct. Based on all of these comments, I'm going to add a bit more detail to try explain it more.

src/alerting-leader.ts
hjon: Noted. I'm going to do both. 🙂

src/types/interfaces.ts
hjon: And when I said "anticipate future use cases", I was attempting to not go overboard. But I also wanted to avoid completely reworking the API if/when we add other interaction types. It may still need adjustment, but I hope it's in the ballpark of what we want so it doesn't feel super jarring (and maybe avoid breaking changes). 🙂

zservies (approved):

maxwellmooney13 (approved):

AidanWZ (approved):

PR #348, opened 2026-04-29
[STREAM-1488 STREAM-1491 STREAM-1498 STREAM-1504] Update axios and follow-redirects

patrickfeeney03:


zservies:
Update changelog.

hjon (changes requested):
As we talked about for another PR, we shouldn't need `follow-redirects` in package.json if it's not a direct dependency. And as Zach said, changelog. 🙂

hjon (approved):

zservies (approved):

PR #356, opened 2026-05-22
[STREAM-1604] - Replace UUID with native alternative

patrickfeeney03:
Build passing in wrapper repo
image


test/helpers/setup-browser-env.js
```diff
@@ -1,2 +1,6 @@
+import { randomUUID } from 'crypto';
+
 global.window = global;
 global.self = global;
+globalThis.crypto = { randomUUID: jest.fn(() => randomUUID())}
```
patrickfeeney03, line 5: Node doesn't have globalThis.crypto... by default so just mock it.

hjon:
We can also remove `@types/uuid`.

hjon (approved):

bruegg (approved):

PR #369, opened 2026-07-10
STREAM-1610 Snyk changes

bruegg:


package.json
```diff
@@ -122,5 +122,10 @@
     "webpack-cli": "^6.0.1",
     "webpack-node-externals": "^2.5.0"
   },
-  "false": {}
+  "false": {},
+  "//overrides": "CVE-2026-41907 (uuid via stanza.io/jxt). CVE-2026-12143 (form-data via superagent). Remove when upstream deps update.",
+  "overrides": {
+    "uuid@<11.1.1": "11.1.1",
```
hjon, line 128: I think this is wrong. We shouldn't have any dependency on `uuid` anymore (except through `devDependencies`, which shouldn't count). Also, `stanza` doesn't have any dependency (direct or transitive) on `uuid`.

PR #371, opened 2026-08-24
[STREAM-2184] Update stanza to v12.22.1

hjon:


tsconfig.json
```diff
@@ -5,7 +5,6 @@
       "declaration": true,
       "esModuleInterop": true,
       "forceConsistentCasingInFileNames": true,
-      "importHelpers": true,
```
hjon, line 8: For this option to work, it seems we were relying on the `tslib` that stanza depended on. Stanza no longer depends uses this option and from what I can tell, I think we should be safe to remove it, as well.

zservies (approved):

AidanWZ (approved):

